PR #7, opened 2021-04-20
Fixed Phone Numbers

uShifty:
Fixes the bug which would allow new players to generate the same phone number as another player in the server.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -593,17 +593,17 @@ QBCore.Player.CreateSerialNumber = function()
 end
 
 QBCore.Player.CreatePhoneNum = function()
-	local UniqueFound = false
-	local number = 0
-	while not UniqueFound do
-		number = "06"..math.random(11111111, 99999999)
-		exports['ghmattimysql']:execute('SELECT * FROM `players` WHERE `charinfo` LIKE "%'.. number ..'%"', false, function(result)
-			if not result[1] then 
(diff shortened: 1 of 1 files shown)


Skillgjii:
tested your edit, and it makes the server crash when people make a new charater

uShifty:
> tested your edit, and it makes the server crash when people make a new charater

I never cared about updating this function due to Kakarot being unsure if he wants to use a wildcard modifier on checking for phone numbers as it gets heavy on larger databases. But this 100% works for me on my server. and 3 other servers I personally implemented this fix into.

vodka-FindOutInRP:
this fix does not work as it causes as it spam the database causing the database to either crash or have such a low response time that causes the server fxserver to crash, adding a wait of 500 could be shorter just used 500 for initial tests seems to resolve the crashes.

QBCorePlayer.CreatePhoneNum = function()
	
	local UniqueFound = false
	local number = 0
	while not UniqueFound do
		number = "06"..math.random(11111111, 99999999)
		QBCore.Functions.ExecuteSql(true,'SELECT citizenid,charinfo FROM `players` WHERE `charinfo` LIKE "%'.. number ..'%"', function(result)
			if not result[1] then 
				UniqueFound = true 
			end
			
		end)
		
		Wait(500)
	end
	return number
end

PR #9, opened 2021-05-20
TP Command, fix for integer coords

emresudo:
Since I couldn't find the toFloat function in Lua, I tried to fix the problem by adding 0.0 as the most convenient. When we enter this integer coordinate, it will prevent the game from breaking.


qbcore-framework:
I know this is super old but what's the problem you were experiencing without this?

emresudo:
Sorry, It would be nice if I add a proof.

Before
image

After
image


I`m using source Fivem Linux Build 3922

PR #14, opened 2021-06-13
Fix Typo on Functions/SpawnVehicle

DmACKGL:


BerkieBb:
This is not the right thing, this ruins it because it was right, there was no typo, look at the function before making changes to it

DmACKGL:
Hey there!
I think I'm blind, can you elaborate?  (Btw, what a way to comment on a VOLUNTARY pull request lul)

BerkieBb:
Still working on my way to react to stuff, sometimes Im being too nice and people dont get that im correcting them so thats why that way but if you look at the function in qb-core, the table returns the heading with the variable a and not h

DmACKGL:
It does? Oh, my bad then. Thanks!

PR #17, opened 2021-06-14
fix(SQL/sanitize): Fixes SQL injections ⚠⚠ MAJOR SQL FUNCTION CHANGE

DmACKGL:
- Added use of parameters instead of just concatenates the data on the query.

### DIFF
--- a/server/events.lua
+++ b/server/events.lua
@@ -237,7 +237,7 @@ AddEventHandler('QBCore:ToggleDuty', function()
 end)
 
 Citizen.CreateThread(function()
-	QBCore.Functions.ExecuteSql(true, "SELECT * FROM `permissions`", function(result)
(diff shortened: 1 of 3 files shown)


DmACKGL:
This commits is on draft due to the need of updating the other resources BEFORE push this.

DmACKGL:
Function Changes: 

"nil" is where the params should be (same has ghmattimysql)
Docs: https://ghmatti.github.io/ghmattimysql/queries/

MySQL.Async.fetchScalar() -> exports['ghmattimysql']:scalar() or QBCore.Functions.ExecuteSql(true, nil,

MySQL.Async.fetchAll() -> exports['ghmattimysql']:execute() or QBCore.Functions.ExecuteSql(true, nil,

MySQL.Async.execute() -> exports['ghmattimysql']:execute() or QBCore.Functions.ExecuteSql(false, nil,

PR #19, opened 2021-06-14
Added /repair command (Admin)

Nahrin-Malakes:


BerkieBb:
Wasn't there already a command named /fix
That does the same, repairing the car.

Issue #21, opened 2021-06-17
[BUG] Money duping


DmACKGL:
At the apartments whenever the player logs off it gets its paycheck.
Event Trace:

https://github.com/qbcore-framework/qb-houses/blob/5bb17eac0dcf9eabfe7bd7d62b0879bf38d9f0f2/server/main.lua#L410
https://github.com/qbcore-framework/qb-core/blob/074dc307f33df4acbea33b317ab0f9d60a8ddb51/server/events.lua#L90
https://github.com/qbcore-framework/qb-core/blob/074dc307f33df4acbea33b317ab0f9d60a8ddb51/server/events.lua#L104

ImpulseFPS:
Just make a loop on the client side to trigger the server event for giving play money every X minutes
and no for some reason it fails gets the player data when the loop is on server side

GhzGarage:
Does it actually reflect onto the players balance upon logout?

ImpulseFPS:
no, because I won't look if the player is logging out the loop will run every 30min for example and it will give him the paycheck
IDK how to explain it better than like this

Player logins into the server
the loop starts running on logging in
after 30min it will trigger the event to give money to the player

ImpulseFPS:
> no, because I won't look if the player is logging out the loop will run every 30min for example and it will give him the paycheck
> IDK how to explain it better than like this
> 
> Player logins into the server
> the loop starts running on logging in
> after 30min it will trigger the event to give money to the player

this way the players won't be able to keep spamming logging out and getting the paycheck

you can do a more advanced version of this loop by checking the last time player got the paycheck and shit but for the sake of it  just an example

Citizen.CreateThread(function()
	while true do
		Citizen.Wait(500 * 3600) -- 30 MIN LOOP FOR PAYCHECK
		if isLoggedIn then
			TriggerServerEvent('ps-paycheck:server:sendPayCheck')
		end
	end
end)

Here is  how i have done it for now i will recode this when i finish other stuff

ImpulseFPS:
I made PR for the fix so you can check that out and just copy the code or you can wait for the PR to be merged with the main branch.

DmACKGL:
Fixed on #22

PR #22, opened 2021-06-17
Money duping exploit fix

ImpulseFPS:
Removed the function trigger to give player money from the logout event

Added new loop on the client-side that will run every 30min and if the player is logged in it will trigger server event for giving the paycheck

Created a new event that will give a paycheck to the player


DmACKGL:
Fixes #21

PR #31, opened 2021-06-22
Update shared.lua

Manliketjb:
updated QBShared.Gangs to load gangs created with qb-gangs

### DIFF
--- a/shared.lua
+++ b/shared.lua
@@ -356,29 +356,7 @@ QBShared.Weapons = {
 }
 
 -- Gangs
-QBShared.Gangs = {
-	["none"] = {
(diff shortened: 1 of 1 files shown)


Manliketjb:
already tested in my server and is required in qb-gangs

MiddleSkillz (changes requested):
There should probably be a check to make sure hangs is present before trying to load it's json. Just do an if GetResourceState('qb-gangs') == 'started' and if its not started just add the no gang to the table instead of trying to load a resource file that doesn't exist. and also if they don't want to use qb-gangs they can add them to the shard manually.

PR #34, opened 2021-06-26
Add logic for manage locales

lunatrik:
add generic function for translation string in QBShared.lua. 
add conf…iguration enviroments Locale


lunatrik (commented):
add functions for manage locale in every qb resource

JericoFX:
Can you provide an actually example, like you need another file right? it.lua or something like that?

lunatrik:
the main change in qb-core is add 2 function in QBShared

QBShared._
QBShared._U

I know this change is currently on hold by BerkieBb. I am continuing with the translations of other scripts. At the moment the shared.lua with all its objects has not been translated. Check pull requests from other scripts like atms / drugs / garages

vanillla1:
But this is not esx
The programmers here love the way it's built qbcore without _U

Mr-Pleb:
I think this may improve the end-user experience by a lot and perhaps ESX servers with different languages will want to transform.

PR #42, opened 2021-07-02
ExecuteSql to ghmattimysql:execute

DmACKGL:


MiddleSkillz (commented):
Is there any benifit to calling the export directly over using the function QB has?

GhzGarage:
Has replacing the function that uses the wait been tested? In some instances I was finding the result was returning too fast so I left the function in solely for the wait. I’m curious if replacing it with the export worked. 

Also Middle, there were inconsistencies with the function when ran through stress tests that the export did not show

PR #53, opened 2021-07-12
Get only stress.

DmACKGL:
No need to get the whole playerdata object.
Needed for: https://github.com/GhzGarage/cosmo_hud/pull/2

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -9,6 +9,11 @@ QBCore.Functions.GetPlayerData = function(cb) -- QBCore.Functions.GetPlayerData(
     end
 end
 
+-- Exports
+function getStress()
+	return QBCore.PlayerData.metadata["stress"]
+end
+
 QBCore.Functions.DrawText = function(x, y, width, height, scale, r, g, b, a, text)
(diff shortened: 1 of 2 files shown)


BerkieBb:
This will not work, the function will return nil.
You're trying to get the metadata stress of QBCore, but QBCore is not a player so it can't get the data, the correct method for this would be through the server by getting the player with the funtion QBCore.Functions.GetPlayer(source) and assigning it to the variable Player and then returning Player.PlayerData.metadata["stress"] or in the client way QBCore.Functions.GetPlayerData().stress

DmACKGL:
Hey @BerkieBb!

I've testest this PR a bunch of times and it works correctly. Actually, I'm just creating a "shortcut" to these lines and getting just the stress metadata instead of the whole object
https://github.com/qbcore-framework/qb-core/blob/f7dc6706baf0ff4361aa5885b3601b792f24f77c/client/functions.lua#L4-L10

GhzGarage:
No longer relevant because cosmo_hud isn't used

PR #63, opened 2021-07-25
add drawtext2d

xenfovn:
add drawtext2d


BerkieBb:
This function already exists under a different name, it's called QBCore.Functions.DrawText

PR #75, opened 2021-07-29
fix(player): Patch exploits with player login / delete

thelindat:
Needs review and testing as I don't have a qb server.

I'll leave it to you to setup appropriate logging and a ban trigger.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -20,7 +20,7 @@ QBCore.Player.Login = function(source, citizenid, newData)
 						PlayerData.gang = {}
 					end
 					local license = QBCore.Functions.GetIdentifier(source, 'license')
-					PlayerData.license ~= license then
+					if PlayerData.license ~= license then
 						DropPlayer(source, 'License mismatch\nThe given citizen id is registered to another Rockstar License')
 						-- add logging for suspicious behaviour
 						return false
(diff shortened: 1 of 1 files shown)


flip4fzL:
I am testing this now with my live city.

BerkieBb:
Looks good to me so far beside there is a missing 'if' on line 23 and you can replace 'QBCore.Functions.GetIdentifier(source, 'license')' on line 501 with the license variable

BerkieBb:
@flip4fzL have you gotten to test this?

flip4fzL:
I wasn't having any issues with it, until server side SQL's were re-written. I had to revert so that my players could re-connect, I will have to do additional testing on my back up dedicated.

thelindat:
> 
> 
> I wasn't having any issues with it, until server side SQL's were re-written. I had to revert so that my players could re-connect, I will have to do additional testing on my back up dedicated.

There was a missing if statement, not sure if you ran it as it was. Problem with how it's written is you could send any citizen id through you wanted and there's no ownership check, plus no prepared statements. Just using NUI devtools you could send through anything and even destroy somebodies database. Granted there could be a check somewhere against that, but I didn't see it on the server.

BerkieBb:
Tested it, I got an infinite loading screen each time so I moved the `return true` part back to where you removed it from and removed all the other `return true`'s and that suprisingly fixed, made sure it always does return true when it's supposed to and it still does

GhzGarage:
Included in latest push

PR #83, opened 2021-08-02
sql file format

CCasusensa:


BerkieBb:
I see this sql file is irrelevant and might be removed, please move your changes to this file which actually has ALL the sql: https://github.com/qbcore-framework/txAdminRecipe/blob/main/qbcore.sql

PR #88, opened 2021-08-03
Add a helper function GetVehiclesInArea

waseem-h:


BerkieBb:
We removed the entity enumerator with a reason so we're not going to add it back

PR #96, opened 2021-08-04
fix interior & dash colors not saving, also custom tires (lettering)

Developer-Bear:


BerkieBb:
The spacing's a bit off lol

Developer-Bear:
that's weird, in my file it's not lol

PR #112, opened 2021-08-07
New ways to editing datas

negbook:


BerkieBb:
Can you explain the change you made?

negbook:
> Can you explain the change you made?

Load datas from csv

BerkieBb:
I want to know how this would be more useful

negbook:
> I want to know how this would be more useful

It would not be more useful if the datas included too many json.encode datas at this moment.

negbook:
Also editing datas from csv is more friendly for server owner but less friendly for the developer.It's depends on qb-core directions.

BerkieBb:
But can you explain why and how this helps in as much detail as you can?

negbook:
No,I can not.It just an idea, a concept.

PR #114, opened 2021-08-08
add job car

vanillla1:


flip4fzL (changes requested):
Can you go ahead and edit those entries, we don't need the trunkspace or trunkslots. Thank you !

vanillla1 (commented):
yes

PR #123, opened 2021-08-13
Cash as an item

Petits-Pieds:
Convert cash as an item


BerkieBb:
This is a breaking commit and we won't do cash as an item as there are unknown exploits and more stuff with it, it's not advised to do this and we won't

PR #124, opened 2021-08-13
Cash as an item

Petits-Pieds:
Convert money has an item, no more database request


BerkieBb:
As I already said in your other pull request, this is highly disadvised because of exploits and other unknown things still wrong with it, I wanted to do it to and got all the messages about what will be wrong, this needs heavy testing before it can get implemented

BerkieBb:
And this was a breaking commit as you changed stuff back/broke stuff in it and removed/renamed stuff which could also break stuff

PR #131, opened 2021-08-16
Updated correct log webhook category.

Bigbearbeasttv:
Updated correct log webhook category.


BerkieBb:
That category doesn't exist

PR #142, opened 2021-08-18
Remove backticks for consistency

waseem-h:


izMystic:
-1 for me, backticks get the weapon hash

BerkieBb:
backticks are needed to get the table easier and for the cars it would be breaking to remove them as the resources that use the name of the vehicle need the hash from there

PR #150, opened 2021-08-21
feat(server/player): PlayerLoaded event for third party resources

TasoOneAsia:
Third party resources like NPWD need the Player instance as soon as available for internal player map handling. This establishes parity with ESX's player load event.


BerkieBb:
The event QBCore:Client:OnPlayerLoaded and QBCore:Server:OnPlayerLoaded gets triggered when the player spawns in qb-spawn/client/main.lua already so this is unnecessary

TasoOneAsia:
@BerkieBb and this isn't a client side event? This is server side emitted

BerkieBb:
Oh sorry, I read it wrong

PR #153, opened 2021-08-23
add duty for /job

vanillla1:


BerkieBb:
You need to add two dots after `Player.PlayerData.job.grade.name` otherwise it'll cause an error

vanillla1:
@BerkieBb fix

PR #154, opened 2021-08-23
add Invalid job or job grade

vanillla1:


BerkieBb:
This is already in the setjob function but there is only no notification so this is an unnecessary double check, consider changing the self.Functions.SetJob command in qb-core/server/player.lua

PR #163, opened 2021-08-28
Add return "closestDistance" for ClosestVehicle

Eralpozcan:
QBCore.Functions.GetClosestVehicle 
With this small addition, we can operate according to the distance of the vehicle in the same function. We are returning a value that already exists in it.

Please let me know if there is a better solution.

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -158,7 +158,7 @@ QBCore.Functions.GetClosestVehicle = function(coords)
 			closestDistance = distance
 		end
 	end
-	return closestVehicle
+	return closestVehicle,closestDistance
 end
 
 QBCore.Functions.GetClosestPed = function(coords, ignoreList) 


Eralpozcan:
The vehicle at a distance was scanned.

Example Using :
`        local checkVehicle = false
        local vehicleCheck, distance = QBCore.Functions.GetClosestVehicle(coords)
        if distance < 4 then 
            if DoesEntityExist(vehicleCheck) then
                checkVehicle = true
            else
                checkVehicle = false
            end
        end
        if not checkVehicle then
            QBCore.Functions.SpawnVehicle(vehicle, function(veh)
                SetVehicleNumberPlateText(veh, "LIFE"..tostring(math.random(1000, 9999)))
                SetVehicleLivery(veh, 1) -- Ambulance Livery
                exports['LegacyFuel']:SetFuel(veh, 100.0)
                TaskWarpPedIntoVehicle(ped, veh, 1)
                TriggerEvent("vehiclekeys:client:SetOwner", GetVehicleNumberPlateText(veh))
                SetVehicleEngineOn(veh, true, true)
            end, coords, true)
        else
            QBCore.Functions.Notify("HeliPad Full", "error")
        end`

evanillaa:
and can add
return closestplayers, closestdistance, closestcoords

Eralpozcan:
I know there is a function for closestPlayers

`QBCore.Functions.GetClosestPlayer`
or
`QBCore.Functions.GetPlayers`

PR #173, opened 2021-08-31
add 	QBCore.Players[source].Functions.Save()

Eralpozcan:
Thanks to this addition, we save when the player exits the game. This fixes some duplication issues.
And when we exit the game, we save everything.

Resolves duplication issues when making changes to his/her/player inventory and outputting it openly.


BerkieBb:
It already does this with the `playerDropped` event, but there is something wrong with that because it fails sometimes

Eralpozcan:
Yes, I agree. According to my tests, if the player exits from F8, the part in playerdropped does not work. It seems to be working if it drops in any way. Before I added this it was not saving on F8 exits.

PR #179, opened 2021-09-02
Notify if inventory is full

davidecab:


server/player.lua
```diff
@@ -321,6 +321,8 @@ QBCore.Player.CreatePlayer = function(PlayerData)
 					end
 				end
 			end
+		else
+			TriggerClientEvent('QBCore:Notify', self.PlayerData.source, "Inventory is full!", "error")
```
FreakOutModz, line 325: In my opinion from past public core that have been made in FiveM, correct translate will be `Your inventory is too heavy`

davidecab: Yea seems legit!

FreakOutModz (approved):

PR #182, opened 2021-09-06
Optimize Command Suggestions

TassieRPCx:
Rather than calling 100 client-events inside of a for loop it creates a table of commands and send the table across to the client in 1 event.

This has resulted in big issues on a server with 50 players hence why I had patched this out.

### DIFF
--- a/server/commands.lua
+++ b/server/commands.lua
@@ -14,12 +14,20 @@ end
 
 QBCore.Commands.Refresh = function(source)
 	local Player = QBCore.Functions.GetPlayer(tonumber(source))
+	local suggestions = {}
 	if Player ~= nil then
 		for command, info in pairs(QBCore.Commands.List) do
 			if QBCore.Functions.HasPermission(source, "god") or QBCore.Functions.HasPermission(source, QBCore.Commands.List[command].permission) then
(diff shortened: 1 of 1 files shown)


TassieRPCx:
Updates the QBCore.Commands.Refresh function and fixes some short falls of it.

server/commands.lua
```diff
@@ -14,12 +14,20 @@ end
 
 QBCore.Commands.Refresh = function(source)
 	local Player = QBCore.Functions.GetPlayer(tonumber(source))
+	local suggestions = {}
 	if Player ~= nil then
 		for command, info in pairs(QBCore.Commands.List) do
 			if QBCore.Functions.HasPermission(source, "god") or QBCore.Functions.HasPermission(source, QBCore.Commands.List[command].permission) then
-				TriggerClientEvent('chat:addSuggestion', source, "/"..command, info.help, info.arguments)
+				--TriggerClientEvent('chat:addSuggestion', source, "/"..command, info.help, info.arguments)
```
FreakOutModz, line 21: This can be remove since is not used

TassieRPCx: yeah correct I forgot to remove it 😂

Issue #218, opened 2021-09-25
[BUG]


SoStressful:
**Describe the bug**
When you update to the latest QBCore update it causes Handcuffing and Sharing of contact details to stop working. There is no script error client-side or server-side.

**To Reproduce**
1. Update to the latest QBCore update
2. Try to share contact details or handcuff someone
3. Look at the top right-hand corner of your screen
4. "No one is nearby"

**Expected behavior**
A successful handcuff or shared contact details.

**Screenshots**
https://i.imgur.com/puNumXr.jpeg

**Questions (please complete the following information):**
 - When you last updated: Every update released.
 - Are you using custom resource? N/A
 - Have you renamed `qb-` to something custom? No

**Additional context**
Tested on someone else's server that also updated and received the same outcome.

lildiznick:
Also get the same "No one Nearby" message when jailing through radial menu.

googleoblivion:
![image](https://user-images.githubusercontent.com/2222973/134770478-35a18540-340a-4d2e-a420-0bf099fad895.png)
I don't know the difference in these styles of creating a table. Reverting back this one line solves this bug.

GhzGarage:
Thanks for the report I will fix this as soon as possible

GhzGarage:
Try this https://github.com/qbcore-framework/qb-core/commit/ddc396d2c99799c60dd1cf94f7f659177ededb90

PR #221, opened 2021-09-25
add return closetDistance

NEvermeant77:
i guess maybe you missed 1 return?

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -190,7 +190,7 @@ function QBCore.Functions.GetClosestPed()
             closestDistance = distance
         end
     end
-    return closestPed
+    return closestPed, closestDistance
 end
 
 -- Vehicle


NEvermeant77:
example:
in qb-ambulance there is some function require the closetdistance
local closestPed, closestDistance = QBCore.Functions.GetClosestPed(coords, PlayerPeds)

Eralpozcan:
I guess they don't find it right to add this. It's obviously causing a problem. Because I opened the same PR before, today it was closed and not added. #163

NEvermeant77:
i see thank man

GhzGarage:
I will look into this. All previous issues and pr’s were closed before this last big update because a lot of code structure changed

Eralpozcan:
@GhzGarage Then if the same problems remain with the current code structure, should we open PR again?

GhzGarage:
Yes of course

GhzGarage:
Also @NEvermeant77 have you tested this pull request or just guessing that that variable should be returned?

Eralpozcan:
@GhzGarage I will tested. We can process according to the distance range. So it's working.

Basicly Example:

```
local closestPed, closestDistance = QBCore.Functions.GetClosestPed(coords, PlayerPeds)
if closestDistance < 5 then
 //bla bla bla whatever you want to do to the one near you.
else
//There is no one near.
end
```

I think this also solves #218 
But I haven't tried that for this problem. I just used it in police job before.

Here we can get reference : [ESX Docs](https://esx-framework.github.io/es_extended/client/functions/game/getclosestplayer/)

PR #222, opened 2021-09-25
closestDistance for ClosestVehicle

Eralpozcan:
#163 Since the rest of the situation here is still not fix.
Here we can check if the vehicle is around that location. 

The function calculates but does not return it to us. It's actually a simple addition to get it to return.

**---------------------------------------------------------**

**Describe Pull request**

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [yes/no] (Be honest) **YES**
- Does your code fit the style guidelines? [yes/no] **YES**
- Does your PR fit the contribution guidelines? [yes/no] **YES**

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -172,7 +172,7 @@ function QBCore.Functions.GetClosestObject()
             closestDistance = distance
         end
     end
-    return closestObject
+    return closestObject,closestDistance
 end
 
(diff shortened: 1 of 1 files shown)


GhzGarage:
Can you add it for lines 175 and 193 as well

Eralpozcan:
Yes of course. I added it now.

PR #227, opened 2021-09-26
fix closestObject/closestVehicle coords

Eralpozcan:
In the previous version we could check the coordinates. In this, we automatically get the coordinate of the player. I hadn't noticed this yesterday when I was testing it. I am attaching this fix. Available in before "Major update" version.

I saw that this caused bugs in some scripts today. So I am posting this update it solves the problem.

**_----------------------_**
**Describe Pull request**

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [yes/no] (Be honest) **YES**
- Does your code fit the style guidelines? [yes/no] **YES**
- Does your PR fit the contribution guidelines? [yes/no] **YES**

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -142,9 +142,7 @@ end
 
 function QBCore.Functions.GetClosestPlayer(coords)
     local ped = PlayerPedId()
-    if coords == nil then
-        coords = GetEntityCoords(ped)
-	end
+    local coords = coords or GetEntityCoords(ped)
     local closestPlayers = QBCore.Functions.GetPlayersFromCoords(coords)
     local closestDistance = -1
     local closestPlayer = -1
@@ -183,10 +181,7 @@ function QBCore.Functions.GetClosestVehicle(coords)
 	local vehicles        = GetGamePool('CVehicle')
 	local closestDistance = -1
 	local closestVehicle  = -1
-	local coords          = coords
-	if coords == nil then
-		coords = GetEntityCoords(ped)
-	end
+	local coords = coords or GetEntityCoords(ped)
 	for i=1, #vehicles, 1 do
 		local vehicleCoords = GetEntityCoords(vehicles[i])
 		local distance = #(vehicleCoords - coords)
@@ -196,7 +191,7 @@ function QBCore.Functions.GetClosestVehicle(coords)
 			closestDistance = distance
 		end
 	end
-	return closestVehicle,closestDistance
+	return closestVehicle, closestDistance
 end
 
 function QBCore.Functions.GetClosestObject(coords)
@@ -204,10 +199,7 @@ function QBCore.Functions.GetClosestObject(coords)
     local objects = GetGamePool('CObject')
     local closestDistance = -1
     local closestObject = -1
-    local coords          = coords
-	if coords == nil then
-		coords = GetEntityCoords(ped)
-	end
+    local coords = coords or GetEntityCoords(ped)
     for i = 1, #objects, 1 do
         local objectCoords = GetEntityCoords(objects[i])
         local distance = #(objectCoords - coords)
@@ -216,7 +208,7 @@ function QBCore.Functions.GetClosestObject(coords)
             closestDistance = distance
         end
     end
-    return closestObject,closestDistance
+    return closestObject, closestDistance
 end
 
 -- Vehicle
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -177,29 +177,37 @@ function QBCore.Functions.GetPlayersFromCoords(coords, distance)
     return closePlayers
 end
 
-function QBCore.Functions.GetClosestVehicle()
-    local ped = PlayerPedId()
-    local coords = GetEntityCoords(ped)
-    local vehicles = GetGamePool('CVehicle')
-    local closestDistance = -1
-    local closestVehicle = -1
-    for i = 1, #vehicles, 1 do
-        local vehicleCoords = GetEntityCoords(vehicles[i])
-        local distance = #(vehicleCoords - coords)
-        if closestDistance == -1 or closestDistance > distance then
-            closestVehicle = vehicles[i]
-            closestDistance = distance
-        end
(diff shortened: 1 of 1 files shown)


GhzGarage:
The coords being passed as a parameter would be the exact same coords that are grabbed when these functions are called so what is the point of checking if the parameter is nil?

Eralpozcan:
I actually want the player to use its coordinate if I'm not sending parameters. If I am sending a parameter I want it to use the value I am sending. It was like this in the previous version. In the current version, it uses the player's position directly. This prevents me from scanning in a certain area.
In its current state, I'm just doing a distance check around the player.I hope there is nothing I missed eye.

GhzGarage:
It’s already setup to use the players coordinates. The only purpose of this PR is to check for the closest something at specific coordinates. 

Do you have an example of why you need to check for something at coords that aren’t the players? That’s kinda the whole purpose of getting things closest to the player

Eralpozcan:
This is a different spawn point of the vehicle, so it is not close to the player. In this way, I check if there is any vehicle in that specific coordinate. And accordingly, I bring the vehicle there.
If there is a vehicle, I do not bring a vehicle in that area so that vehicles do not come on top of each other.


_If we do not do this, the vehicles will spawn one after the other. And also, I use camera mode in a script I have made, when I buy it, it is in a different position and I have to remove the vehicle that it looked like in that catalog._

Full Code :
```
RegisterNetEvent("qb-ambulancejob:client:HeliTakeOut")
AddEventHandler("qb-ambulancejob:client:HeliTakeOut",function(data)
    local vehicle = nil
    local coords = vector3(351.85, -588.22, 74.16)
    if QBCore.Functions.GetPlayerData().metadata["licences"]["plane"] then
        if data.type == "maverick" then
            vehicle = "polmav"
        elseif data.type == "mh65" then
            QBCore.Functions.Notify("Cant use", "error")
        elseif data.type == "aw101" then
            QBCore.Functions.Notify("Cant use", "error")
        end
        
        local checkVehicle = false
        local vehicleCheck, distance = QBCore.Functions.GetClosestVehicle(coords)
        if distance < 4 then 
            if DoesEntityExist(vehicleCheck) then
                checkVehicle = true
            else
                checkVehicle = false
            end
        end
        if not checkVehicle then
            QBCore.Functions.SpawnVehicle(vehicle, function(veh)
                SetVehicleNumberPlateText(veh, "LIFE"..tostring(math.random(1000, 9999)))
                SetVehicleLivery(veh, 1) -- Ambulance Livery
                exports['LegacyFuel']:SetFuel(veh, 100.0)
                ClearUI()
                TaskWarpPedIntoVehicle(ped, veh, 1)
                TriggerEvent("vehiclekeys:client:SetOwner", GetVehicleNumberPlateText(veh))
                SetVehicleEngineOn(veh, true, true)
            end, coords, true)
        else
            QBCore.Functions.Notify("HeliPad Full", "error")
        end
    else
       QBCore.Functions.Notify("No Have License", "error")
    end
end)`

File: client/functions.lua [hunk 1]
```diff
@@ -177,29 +178,36 @@ function QBCore.Functions.GetPlayersFromCoords(coords, distance)
...
-    return closestVehicle, closestDistance
+function QBCore.Functions.GetClosestVehicle(coords)
+	local ped = PlayerPedId()
+	local vehicles        = GetGamePool('CVehicle')
+	local closestDistance = -1
+	local closestVehicle  = -1
+	local coords          = coords
+	if coords == nil then
+		coords = GetEntityCoords(ped)
+	end
```
AvarianKnight, line 189: This can be simplified a lot with
```lua
local coords = coords or GetEntityCoords(ped)
```

client/functions.lua
```diff
@@ -177,29 +178,36 @@ function QBCore.Functions.GetPlayersFromCoords(coords, distance)
...
 end
 
-function QBCore.Functions.GetClosestObject()
+function QBCore.Functions.GetClosestObject(coords)
     local ped = PlayerPedId()
-    local coords = GetEntityCoords(ped)
     local objects = GetGamePool('CObject')
     local closestDistance = -1
     local closestObject = -1
+    local coords          = coords
```
AvarianKnight, line 207: Same here

client/functions.lua
```diff
@@ -208,7 +216,7 @@ function QBCore.Functions.GetClosestObject()
             closestDistance = distance
         end
     end
-    return closestObject, closestDistance
+    return closestObject,closestDistance
```
AvarianKnight, line 219: Why randomly remove a white space :(

client/functions.lua
```diff
@@ -177,29 +178,36 @@ function QBCore.Functions.GetPlayersFromCoords(coords, distance)
...
-        local distance = #(vehicleCoords - coords)
-        if closestDistance == -1 or closestDistance > distance then
-            closestVehicle = vehicles[i]
-            closestDistance = distance
-        end
-    end
-    return closestVehicle, closestDistance
+function QBCore.Functions.GetClosestVehicle(coords)
+	local ped = PlayerPedId()
+	local vehicles        = GetGamePool('CVehicle')
```
AvarianKnight, line 183: Would probably be a good idea to keep the code consistency with how this was before, this doesn't really improve readability at all

AvarianKnight (approved):

AvarianKnight (changes requested):
I accidentally clicked approve

client/functions.lua
Eralpozcan: Did it work for some reason, it looked bad to me :D

client/functions.lua
Eralpozcan: Actually, I want to eliminate the "or" possibility by making a check here. If it works stably, of course I add it as "or".

Maybe it's better to do it this way.
`if not coords then coords = GetEntityCoords(ped) end`

Eralpozcan:
I took a video to see its usage better.

[Streamable](https://streamable.com/kk1jk5)

PR #232, opened 2021-09-27
feat(server/functions): Add GetQBPlayers() fn

TasoOneAsia:
**Describe Pull request**
A wrapper that returns an array of QB Player instances. Mainly to provide a better solution to the GetPlayers() -> GetPlayer pattern seen everywhere.

Common Pattern Previous to FN:
```lua
local players = QBCore.Functions.GetPlayers()

for _,v in ipairs(players) do
  local qbPlayer = QBCore.Functions.GetPlayer(v)
  
  if qbPlayer.PlayerInfo.WHATEVER == 'nice' then
    -- Target certain property of player instance
  end
end
```

Pattern with FN:
```lua
local players = QBCore.Functions.GetQBPlayers()

for _,v in ipairs(players) do
  if v.PlayerInfo.WHATEVER == 'nice' then
    -- Target certain property of player instance
  end
end
```

Reduces the amount of cross resource function ref calls significantly when iterating.

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [yes/no] (Be honest) ✔️ 
- Does your code fit the style guidelines? [yes/no]  ✔️
- Does your PR fit the contribution guidelines? [yes/no] ✔️

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -71,6 +71,11 @@ function QBCore.Functions.GetPlayers()
 	return sources
 end
 
+-- Will return an array of QB Player class instances
+-- unlike the GetPlayers() wrapper which only returns IDs
+function QBCore.Functions.GetQBPlayers()
+	return QBCore.Players
+end
 -- Paychecks (standalone - don't touch)
 
 function PaycheckLoop()


TasoOneAsia:
Note: This doesn't need to copy the table before being sent (as a protection against direct mutation) as serialization over events will  copy all basic types aside from function refs.

PR #246, opened 2021-10-01
feat (Functions/Math): Added Precision Rounding so that people can round to nearest decimal efficiently.

IdrisDose:
**Describe Pull request**
So the Lua native math library does not seem to include rounding to decimal functionality e.g; math.round(0.02314, 2), I implemented a function to handle that.

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? ✅
- Does your code fit the style guidelines? ✅
- Does your PR fit the contribution guidelines? ✅

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -451,3 +451,22 @@ function QBCore.Functions.SetVehicleProperties(vehicle, props)
 		end
 	end
 end
+
+-- Extra Math Functions
+
+-- Math Rounding Credits: http://lua-users.org/wiki/SimpleRound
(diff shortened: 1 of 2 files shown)


Eralpozcan:
It seems to do it when I check the document. Is there a spot we haven't seen? Can you show with an example?

[Lua-Round](http://lua-users.org/wiki/SimpleRound)

IdrisDose:
If you look, there is no math.round, the round functions specified in that link you sent is the discussion on how to implement it, I simply copied the latest post on that page as per the comments in the code.

PR #255, opened 2021-10-07
feat(notify): Config to customise behaviour of notification

LiamDormon:
**Describe Pull request**
Adds QBCore.Config.Notify to customise the behaviour of notifications

- Does your code fit the style guidelines? probably
- Does your PR fit the contribution guidelines? probably

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -58,6 +58,10 @@ function QBCore.Functions.DrawText3D(x, y, z, text) -- Use local function instea
     ClearDrawOrigin()
 end
 
+RegisterNUICallback("GetNotifyConfig", function(data, cb)
+	cb(json.encode(QBCore.Config.Notify))
+end)
+
 function QBCore.Functions.Notify(text, textype, length)
 	if type(text) == "table" then
 		local ttext = text.text or 'Placeholder'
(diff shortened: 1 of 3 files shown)


JericoFX:
you can simplify the code like this 

const { useQuasar } = Quasar;
const { onMounted, onUnmounted,ref } = Vue;
const app = Vue.createApp({
  setup() {
    const NotificationConfig = ref({
      stack:false,
      position:'top-right',
      progress: true
    })
    const $q = useQuasar();

    const showNotif = (e) => {
   
      const text = e.data.text;
      const length = e.data.length;
      const type = e.data.type;
      const caption = e.data.caption;
      NotificationConfig.value = e.data.NotificationConfig
   
      switch (type) {
        case 'success':
          color = 'green';
          icon = 'done';
          break;
        case 'primary':
          color = 'blue';
          icon = 'info';
          break;
        case 'error':
          color = 'red';
          icon = 'dangerous';
          break;
        case 'police':
          color = 'blue';
          icon = 'local_police';
          break;
        case 'ambulance':
          color = 'red';
          icon = 'fas fa-ambulance';
          break;
      }

      if (text.length > 100) {
        multiline = true;
      } else {
        multiline = false;
      }

      $q.notify({
        message: text,
        caption: caption,
        multiLine: multiline,
        color: color,
        group: NotificationConfig.value.stack,
        progress: NotificationConfig.value.progress,
        position: NotificationConfig.value.position,
        timeout: length,
        icon: icon,
      });
    };
    onMounted(() => {
      window.addEventListener('message', showNotif);
    });
    onUnmounted(() => {
      window.removeEventListener('message', showNotif);
    });
    return {};
  },
});
app.use(Quasar);
app.mount('#q-app');

PR #277, opened 2021-10-17
Optimize Load Inventory function

CxAUS:
Change from * which selects every column in the players table to just selecting the inventory column.

This should result in slightly faster loading times.

**Describe Pull request**
This PR results in a better loading time when fetching the inventory from the Players table

I think correctly constructed MySQL Queries are the bigger performance increases rather then avoiding the problem and switching to OxMYSQL as badly written queries will only be slightly faster.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -525,7 +525,7 @@ end
 
 QBCore.Player.LoadInventory = function(PlayerData)
     PlayerData.items = {}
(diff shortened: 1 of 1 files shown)


CxAUS:
https://github.com/qbcore-framework/qb-core/compare/main...CxAUS:patch-3

GhzGarage:
Closed due to no response

PR #307, opened 2021-11-03
Update Shared.lua

TrueSunnyD:
**Describe Pull request**
Added vehicles in the Sports category and Super category for the luxury shop.

only need to change the Shared.lua

This is my first time doing this, so I hope I did it right lol


zfbx:
This is not how you would do this, this overwrites the previous values of these cars that are already in the sharred.lua, you'll want to just change them from "pdm" to "luxury" on your local setup if you want them there instead

TrueSunnyD:
> This is not how you would do this, this overwrites the previous values of these cars that are already in the sharred.lua, you'll want to just change them from "pdm" to "luxury" on your local setup if you want them there instead

Ahhh okay that makes sense! My thought process on this was to make it to where people would still be able to have the option of being able to purchase the vehicles through PDM as well as luxury, that's why I didn't delete the originals and added them on being in the Luxury Shop. 

Does that still "overwrite" them? or does that work? or are you able to add 2 shop categories to one car? 

(still learning the coding ways, so I apologize lol)

GhzGarage:
Right now it’s just one or the other

zfbx:
good to close

Mojito-Fivem:
> Does that still "overwrite" them? or does that work? or are you able to add 2 shop categories to one car?
Yes, they share the same key in the table so they will overwrite.

TrueSunnyD:
Ok This is good to close, I will re-write another one :)

Mojito-Fivem:
@TrueSunnyD it may be interesting if you can get the `shop` property to take a table argument to allow for multiple shops, this will take some amout of work in qb-vehicleshop

PR #311, opened 2021-11-07
Refactor functions

Mojito-Fivem:
This pull request refactors client/functions.lua for the following reasons:

1) There is no reason why a local round function should exist when such a function exists in the QBCore.Shared namespace
2) I see no reason why Trim can't be a global utility function in QBCore.Shared

This should make these functions more uniform with the rest of qb-core.

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -265,14 +265,6 @@ end
 
 -- Vehicle
 
-local function Trim(value)
-    if value then
-        return (string.gsub(value, '^%s*(.-)%s*$', '%1'))
-    else
-        return nil
-    end
-end
-
 function QBCore.Functions.SpawnVehicle(model, cb, coords, isnetworked)
     local model = GetHashKey(model)
     local ped = PlayerPedId()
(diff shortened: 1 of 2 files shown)


GhzGarage:
I'm good with this i just wanna make sure that round shared function returns the same value as the local function

roobr:
Hey @Mojito-Fivem 

This trim function can cause increased CPU time when used in loops FYI :) Kak came across it

darknight2590:
After the last update, there is a high consumption, in addition to In addition to a continuous decrease fps and increase ms

roobr:
There is an update that removed the getplate from being used in loops and returned local Trim functions to anticheat and such that use them in loops.

This Getplate is an optional thing.

Mojito-Fivem:
> Hey @Mojito-Fivem
> 
> This trim function can cause increased CPU time when used in loops FYI :) Kak came across it

The difference is negligible, micro optimizations really. Readability is more important imo and having it in the global QBCore.Shared namespace makes it more versatile.

If you *really* want to, theres nothing stopping you from caching this function to a local variable in your loops.

i.e.
```lua
CreateThread(function()
    local DrawText3D = QBCore.Functions.DrawText3D -- cache to local variable
    while true do
        DrawText3D(...) -- used cached function in while loop
    end
end)
```

Ultimately you're only gaining microseconds per tick so it's up to you to pick readability over pedantic micro optimisation.

Mojito-Fivem:
> After the last update, there is a high consumption, in addition to In addition to a continuous decrease fps and increase ms He was fps Good but now is bad

Doubt. Run a profiler and provide accurate data not your own narratives please.

PR #328, opened 2021-11-14
Sort items and delete unused

evanillaa:
**Describe Pull request**
Sort items and delete unused

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [yes]


jnccloud (commented):
I wanted to say that removing items like puremeth, acetone, and other drug related items (not positive on everything that was removed here, but I already noticed these) could be a breaking change for existing servers. A lot of servers have put in the time and work to make these items usable for their criminal economy. Removing items is one of the reasons I always have to disregard changes to the items table in shared.lua when I update. Just my two cents!

evanillaa:
These items are not available in the project whose scripts have been deleted

PR #331, opened 2021-11-14
feat(client/functions): GetPlayersInArea

CherryozZ:


GhzGarage:
The coords needs changed because you will already be passing a vector 3. You should instead check if coords exist and if not then get the players coords.

client/functions.lua
```diff
@@ -134,6 +134,22 @@ function QBCore.Functions.GetPlayers()
     return GetActivePlayers()
 end
 
+function QBCore.Functions.GetPlayersInArea(coords, area)
+    local players, playersInArea = QBCore.Functions.GetPlayers(), {}
+    coords = vector3(coords.x, coords.y, coords.z)
```
Mojito-Fivem, line 139: https://github.com/qbcore-framework/qb-core/blob/f41e23b69a604d0d4ad35dfff7c59b44611d6f7e/client/functions.lua#L157 this should match the type checking used in other functions as @GhzGarage mentioned.

Mojito-Fivem (changes requested):

Issue #345, opened 2021-11-28
[SUGGESTION] PVP (Enabled/Disabled) Option


Duguayster:
**Is your feature request related to a problem? Please describe.**
Kind of, I don't like being shot in the face randomly, and I'm sure I'm not alone here.

**Describe the feature you'd like**
Nice and simple. Some servers like to have it as a PVE to where you just chill, enjoy the systems in place and do your own thing without having the fear of being randomly shot in the face at any moment.

**Describe alternatives you've considered**
Unless you want to put a bunch of other systems in place and rules that are just downright ridiculous to curb this without having an actual PVE option, would be just nice to be able to toggle it.

**Additional context**
Thank you for your consideration.

Mitroxs:
So disable server wide PVP or a option for each player?

Duguayster:
Server wide, making a PVE server while enjoying the QBCore platform.

Just a toggleable option in the config would be cool.

I found the friendly fire code in the client/events.lua and got pvp turned off. Just thought be a nice feature to have.

Mitroxs:
> Server wide, making a PVE server while enjoying the QBCore platform.
> 
> Just a toggleable option in the config would be cool.
> 
> I found the friendly fire code in the client/events.lua and got pvp turned off. Just thought be a nice feature to have.
#352

Mitroxs:
@GhzGarage Someone close this as my last PR was related to this issue

JajarGaming:
> **Is your feature request related to a problem? Please describe.** Kind of, I don't like being shot in the face randomly, and I'm sure I'm not alone here.
> 
> **Describe the feature you'd like** Nice and simple. Some servers like to have it as a PVE to where you just chill, enjoy the systems in place and do your own thing without having the fear of being randomly shot in the face at any moment.
> 
> **Describe alternatives you've considered** Unless you want to put a bunch of other systems in place and rules that are just downright ridiculous to curb this without having an actual PVE option, would be just nice to be able to toggle it.
> 
> **Additional context** Thank you for your consideration.

Close this "Issue"

PR #352, opened 2021-12-01
Introduces toggle option for PVP

Mitroxs:
**Describe Pull request**
https://github.com/qbcore-framework/qb-core/issues/345

Don't believe its necessary in this case
Yes
From what i have read and the way i understood the guidelines, yes


Mojito-Fivem (approved):
LGTM

Mitroxs:
The command is an additional feature, as i can see it being nice for events etc, without having to restart the server.

Reason for creating the extra event client side is incase there are players on the server while the command is being called. And its purpose is to set the state for those aswell.

PR #358, opened 2021-12-03
Introduction to Routingbuckets

Mitroxs:
**Describe Pull request**
From what i have seen, there is no real implementation for routingbuckets in the current state of qb-core.
This PR contains all the functions and utilities for easy usage of routingbuckets.

No

Yes
It should, let me know if it does not and i will fix that
Yes

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -107,6 +107,75 @@ function QBCore.Functions.GetDutyCount(job)
     return count
 end
 
+--- Routingbucket stuff (Only touch if you know what you are doing)
+_G.QBCore.Player_Buckets = {} -- Bucket array containing all players that have been set to a different bucket
+_G.QBCore.Entity_Buckets = {} -- Bucket array containing all entities that have been set to a different bucket
+
+--- Will set the provided player id / source into the provided bucket id
+function QBCore.Functions.SetPlayerBucket(player_source --[[int]],bucket --[[int]])
+    if player_source and bucket then
+        local plicense = QBCore.Functions.GetIdentifier(player, 'license')
+        SetPlayerRoutingBucket(player_source, bucket)
+        _G.QBCore.Player_Buckets[plicense] = {player_id = player_source, player_bucket = bucket}
+        return true
+    else
+        return false
+    end
+end
+
+--- Will set any entity into the provided bucket, for example peds / vehicles / props / etc...
+function QBCore.Functions.SetEntityBucket(entity --[[int]],bucket --[[int]])
+    if entity and bucket then
+        SetEntityRoutingBucket(entity, bucket)
+        _G.QBCore.Entity_Buckets[entity] = {entity_id = entity, entity_bucket = bucket}
+        return true
+    else
+        return false
+    end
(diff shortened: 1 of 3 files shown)


server/functions.lua
```diff
@@ -107,6 +107,75 @@ function QBCore.Functions.GetDutyCount(job)
...
+end
+
+--- Will return true / false wheter the mentioned player id is present in the bucket provided
+function QBCore.Functions.IsPlayerInBucket(player_source --[[int]] ,bucket --[[int]])
+    local curr_player_bucket = GetPlayerRoutingBucket(player_source)
+    if curr_player_bucket == bucket then
+        return true
+    else
+        return false
+    end
```
TasoOneAsia, line 176: Can be simplified slightly.

```suggestion
    return curr_player_bucket == bucket
```

server/functions.lua
```diff
@@ -107,6 +107,75 @@ function QBCore.Functions.GetDutyCount(job)
     return count
 end
 
+--- Routingbucket stuff (Only touch if you know what you are doing)
+_G.QBCore.Player_Buckets = {} -- Bucket array containing all players that have been set to a different bucket
+_G.QBCore.Entity_Buckets = {} -- Bucket array containing all entities that have been set to a different bucket
```
TasoOneAsia, line 112: Whats the reasoning for adding these to the shared object? For this data to be up to date across all consuming resources, a resource would need to refresh the object before operations. It may be better to expose this data through some getter function that would copy the entity table state at runtime.

Also these data structures may be more efficient as a Map using entity ID as the key.

TasoOneAsia (changes requested):
Couple things to note and think about. Depending on response to my feedback, I can provide further pointers.

server/functions.lua
Mitroxs: Yeah, true, been a long time since i touched lua properly so kind of forgot about this

server/functions.lua
Mitroxs: Thanks for your taking your time to write the feedback btw, also from looking at the QBcore object, it made the most sense to add them there. Regarding the datastructure i was not so sure how to do it in a good way, so feel free to edit it and add whatever u believe is best

TasoOneAsia:
So in regards to the table used for storing entity buckets, why do we want this state abstracted to QBCore within a wrapper instead of just using the underlying native bucket state? Is there a valid use case you can think of that requires it?

Mitroxs:
> So in regards to the table used for storing entity buckets, why do we want this state abstracted to QBCore within a wrapper instead of just using the underlying native bucket state? Is there a valid use case you can think of that requires it?

From past experience using routingbuckets, i think its fair and necessary in some cases. I cannot come up with a specific use case,

Mitroxs:
Would a similar export to [exports('GetCoreObject'](https://github.com/qbcore-framework/qb-core/blob/e7c456228986950490dcf431c9f3cc99559f528a/server/main.lua#L7-L9) just with the bucket objects, be better?

PR #361, opened 2021-12-03
Updated 'setjob' command to check job validity

the3leven:
Added a validity check to 'setjob' command that returns an error

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -299,3 +299,17 @@ function QBCore.Functions.IsLicenseInUse(license)
     end
     return false
 end
+
+-- Job Check
+
+function QBCore.Functions.DoesJobExist(job, grade)
+    if job and grade then
+        for k, v in pairs(QBCore.Shared.Jobs) do
+            if job == k and v.grades[grade] then
+                return true
+            end
+        end
+    end
+    
+    return false
+end
--- a/server/commands.lua
+++ b/server/commands.lua
@@ -142,7 +142,11 @@ QBCore.Commands.Add('setjob', 'Set A Players Job (Admin Only)', { { name = 'id',
     local src = source
     local Player = QBCore.Functions.GetPlayer(tonumber(args[1]))
     if Player then
-        Player.Functions.SetJob(tostring(args[2]), tonumber(args[3]))
+       if QBCore.Functions.DoesJobExist(args[2], args[3]) then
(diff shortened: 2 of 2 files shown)


server/functions.lua
```diff
@@ -299,3 +299,17 @@ function QBCore.Functions.IsLicenseInUse(license)
     end
     return false
 end
+
+-- Job Check
+
+function QBCore.Functions.DoesJobExist(job, grade)
+    if job and grade then
+        for k, v in pairs(QBCore.Shared.Jobs) do
```
Mojito-Fivem, line 307: Why are you making this O(n) when it could just be O(1) by using direct lookup?

server/functions.lua
```diff
@@ -299,3 +299,17 @@ function QBCore.Functions.IsLicenseInUse(license)
     end
     return false
 end
+
+-- Job Check
+
+function QBCore.Functions.DoesJobExist(job, grade)
```
Mojito-Fivem, line 305: Why does this need to be a separate function?

Mojito-Fivem (changes requested):
What is the purpose of this pull request even?

server/functions.lua
the3leven: You could run it directly inside the command I guess..

server/functions.lua
the3leven: What do you mean?

server/functions.lua
Mojito-Fivem: But you're not using it elsewhere, if you're running _inside_ the function, do just that

server/functions.lua
Mojito-Fivem: I mean this algorithm you've wrote is O(n) for no reason when you can very easily use an O(1) direct lookup of the table

the3leven: If there is a better way that is better "written" you can do as you wish.

>What is the purpose of this pull request even?
I noticed people asking for it in the Suggestions channel and its something I need too for my staff team.
Its called a "request" for a reason , so... :P

Mojito-Fivem: This is your pull request so.... it probably won't get merged if you're using algorithms that don't even need to exist since this can be simplified to a simple direct index of the table. There are code guidelines for a reason to ensure that qbcore contains only quality code.

PR #369, opened 2021-12-08
(add): create blip function

YeahItsHayden:
**Describe Pull request**
This just adds a function for creating blips, as I've been coding further with QBCore, I like to take some shortcuts, and this is something I've included in my local framework that I believe could be beneficial to everyone

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -77,6 +77,16 @@ function QBCore.Functions.CreateBlip(coords, sprite, display, scale, colour, sho
     end
 end
 
+function QBCore.Functions.RequestAnimDict(animDict)
+	if not HasAnimDictLoaded(animDict) then
+		RequestAnimDict(animDict)
+
+		while not HasAnimDictLoaded(animDict) do
+			Wait(4)
+		end
+	end
+end
+
 RegisterNUICallback('getNotifyConfig', function(_, cb)
     cb(QBCore.Config.Notify)
 end)
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -60,6 +60,23 @@ function QBCore.Functions.DrawText3D(x, y, z, text)
     ClearDrawOrigin()
 end
 
+function QBCore.Functions.CreateBlip(coords, sprite, display, scale, colour, shortRange, title)
+    print(coords, sprite, display, scale, colour, shortRange, title)
+    if not coords or not sprite or not display or not scale or not colour or shortRange == nil or not title then 
+        print("Blip failed to create, most likely missed a setting")
+    else 
+        blip = AddBlipForCoord(coords)
+        SetBlipSprite(blip, sprite)
+        SetBlipDisplay(blip, display)
+        SetBlipScale(blip, scale)
+        SetBlipColour(blip, colour)
+        SetBlipAsShortRange(blip, shortRange)
+        BeginTextCommandSetBlipName("STRING")
+        AddTextComponentString(title)
+        EndTextCommandSetBlipName(blip)
+    end
+end
+
 RegisterNUICallback('getNotifyConfig', function(_, cb)
     cb(QBCore.Config.Notify)
 end)
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -61,9 +61,9 @@ function QBCore.Functions.DrawText3D(x, y, z, text)
 end
 
 function QBCore.Functions.CreateBlip(coords, sprite, display, scale, colour, shortRange, title)
-    print(coords, sprite, display, scale, colour, shortRange, title)
     if not coords or not sprite or not display or not scale or not colour or shortRange == nil or not title then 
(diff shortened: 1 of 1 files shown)


YeahItsHayden:
I also made a requestanim dict etc I can add those aswell if you want.

darknight2590 (approved):

client/functions.lua
```diff
@@ -60,6 +60,33 @@ function QBCore.Functions.DrawText3D(x, y, z, text)
     ClearDrawOrigin()
 end
 
+function QBCore.Functions.CreateBlip(coords, sprite, display, scale, colour, shortRange, title)
+    print(coords, sprite, display, scale, colour, shortRange, title)
```
Mojito-Fivem, line 64: print statement

client/functions.lua
```diff
@@ -60,6 +60,33 @@ function QBCore.Functions.DrawText3D(x, y, z, text)
...
+    if not coords or not sprite or not display or not scale or not colour or shortRange == nil or not title then
```
Mojito-Fivem, line 65: What if somebody wants a blip without text? or without a colour etc?

client/functions.lua
```diff
@@ -60,6 +60,33 @@ function QBCore.Functions.DrawText3D(x, y, z, text)
     ClearDrawOrigin()
 end
 
+function QBCore.Functions.CreateBlip(coords, sprite, display, scale, colour, shortRange, title)
+    print(coords, sprite, display, scale, colour, shortRange, title)
+    if not coords or not sprite or not display or not scale or not colour or shortRange == nil or not title then 
+        print("Blip failed to create, most likely missed a setting")
+    else 
+        blip = AddBlipForCoord(coords)
```
Mojito-Fivem, line 68: No consideration for other blip types?

Mojito-Fivem (changes requested):

client/functions.lua
YeahItsHayden: Can you elaborate on blip types?
display is the only pram that really changes the type of blip (IE: How the blip is shown), in which it is taken into consideration

client/functions.lua
YeahItsHayden: If a color pram isn't defined in a blip it'll just default to white, colour pram is required

Title will take an empty string fine, like I said in the PR this is something that I've been using without issues,

I'm taking all blip aspects into consideration so I have to check for everything, but feel free to PR something else as I only pushed this because I thought it'd be useful to others

client/functions.lua
Mojito-Fivem: AddBlipForRadius etc

client/functions.lua
Mojito-Fivem: > If a color pram isn't defined in a blip it'll just default to white
So, it's not required then?

darknight2590:
im using this
```
function QBCore.Functions.CreateBlip(id, data)
    local blip = AddBlipForCoord(data.x, data.y, data.z)

    if data.sprite then SetBlipSprite(blip, data.sprite) end
    if data.range then SetBlipAsShortRange(blip, data.range) else SetBlipAsShortRange(blip, true) end
    if data.color then SetBlipColour(blip, data.color) end
    if data.display then SetBlipDisplay(blip, data.display) end
    if data.playername then SetBlipNameToPlayerName(blip, data.playername) end
    if data.showcone then SetBlipShowCone(blip, data.showcone) end
    if data.secondarycolor then SetBlipSecondaryColour(blip, data.secondarycolor) end
    if data.friend then SetBlipFriend(blip, data.friend) end
    if data.mission then SetBlipAsMissionCreatorBlip(blip, data.mission) end
    if data.route then SetBlipRoute(blip, data.route) end
    if data.friendly then SetBlipAsFriendly(blip, data.friendly) end
    if data.routecolor then SetBlipRouteColour(blip, data.routecolor) end
    if data.scale then SetBlipScale(blip, data.scale) else SetBlipScale(blip, 0.8) end

    BeginTextCommandSetBlipName("STRING")
    AddTextComponentString(data.name)
    EndTextCommandSetBlipName(blip)

    Blips[id] = {blip = blip, data = data}
end

function QBCore.Functions.RemoveBlip(id)
    local blip = Blips[id]
    if blip then RemoveBlip(blip.blip) end
    Blips[id] = nil
end

function QBCore.Functions.HideBlip(id, toggle)
    local blip = Blips[id]
    if not blip then return end
    if toggle then 
        SetBlipAlpha(blip.blip, 0)
        SetBlipHiddenOnLegend(blip.blip, true)
    else
        SetBlipAlpha(blip.blip, 255)
        SetBlipHiddenOnLegend(blip.blip, false)
    end
end

function QBCore.Functions.GetBlip(id)
    local blip = Blips[id]
    if not blip then return false end
    return blip
end
```

PR #377, opened 2021-12-17
Fixes QBCore.Functions.HasItem [Critical](!)

Mojito-Fivem:
**Describe Pull request**
Fixes QBCore.Functions.HasItem, this function does not work and now it does :)

This PR is very important

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -18,13 +18,15 @@ function QBCore.Functions.GetCoords(entity)
 end
 
 function QBCore.Functions.HasItem(item)
+    local p = promise.new()
     QBCore.Functions.TriggerCallback('QBCore:HasItem', function(result)
         if result then
-            return true
+            p:resolve(true)
         end
-        return false
+        p:resolve(false)
     end, item)
-    return false
+
+    return Citizen.Await(p)
 end
 
 -- Utility


LifeGoal:
This does not work when trying to use multiple QBCore.Functions.HasItem. Like if you would need to search for 3 items or whatever.

Mojito-Fivem:
This function is only used to detect the presence of a single item

LifeGoal:
Absolutely, you always check for 1 item. I'm not trying to do something like this
```lua
QBCore.Functions.HasItem("item1", "item2", "item3")
```
or whatever.

What I'm saying is that, it stops working if you try to do something like this:
```lua
local hasItem1 = QBCore.Functions.HasItem("whateverItem1") -- Works for this one
local hasItem2 = QBCore.Functions.HasItem("whateverItem2") -- Here the entire Thread stops (if you are using it inside a thread)
```

Mojito-Fivem:
Have you tried Citizen.Awaiting the result since it returns a promise now

LifeGoal:
> Have you tried Citizen.Awaiting the result since it returns a promise now

That worked :)

ItsVinnyX:
> Absolutely, you always check for 1 item. I'm not trying to do something like this
> 
> ```lua
> QBCore.Functions.HasItem("item1", "item2", "item3")
> ```
> 
> or whatever.
> 
> What I'm saying is that, it stops working if you try to do something like this:
> 
> ```lua
> local hasItem1 = QBCore.Functions.HasItem("whateverItem1") -- Works for this one
> local hasItem2 = QBCore.Functions.HasItem("whateverItem2") -- Here the entire Thread stops (if you are using it inside a thread)
> ```

Having similar issues, I am pretty new to FiveM development. Not sure when await the result.

`if itemOne and itemTwo and (math.random(0, 100) <= 80)) then
                HasKey = true
                QBCore.Functions.Notify("Hotwiring was successful!")
                TriggerEvent("vehiclekeys:client:ToggleEngine")
            else
                HasKey = false
                SetVehicleEngineOn(veh, false, false, true)
                QBCore.Functions.Notify("Hotwire failed!", "error")
            end`

Kills the thread immediately after two items are involved.

Mojito-Fivem:
```
local hasItem1 = Citizen.Await(QBCore.Functions.HasItem("item1"))
```

ItsVinnyX:
image

```lua
local hasItemOne = Citizen.Await(QBCore.Functions.HasItem("advancedlockpick"))
local hasItemTwo = Citizen.Await(QBCore.Functions.HasItem("lockpick"))

            if (hasItemOne and hasItemTwo and (math.random(0, 100) <= 60)) then
                HasKey = true
                QBCore.Functions.Notify("Hotwiring was successful!")
                TriggerEvent("vehiclekeys:client:ToggleEngine")
            else
                HasKey = false
                SetVehicleEngineOn(veh, false, false, true)
                QBCore.Functions.Notify("Hotwire failed!", "error")
            end
```
Not familiar with promise either, doesn't the result return true/false?

Mojito-Fivem:
You can't await it there because the promise has already been resolved

LifeGoal:
> Not familiar with promise either, doesn't the result return true/false?



> You can't await it there because the promise has already been resolved

I think it would be better if you told him where he needs to await it tbh

ItsVinnyX:
Yes! Would help a lot!! As well as be appreciated.

> > You can't await it there because the promise has already been resolved
> 
> I think it would be better if you told him where he needs to await it tbh

PR #392, opened 2021-12-24
update Animation task

darknight2590:
**Describe Pull request**

update Animation task function 

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? maybe
- Does your code fit the style guidelines? yeees
- Does your PR fit the contribution guidelines? ya

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -125,7 +125,7 @@ function QBCore.Functions.LoadAnimDict(animDict)
     end
 end
 
-function QBCore.Functions.Animation:new(iPed, iType, iText, iDuration, iDict, iAnim, iFlag)
+function QBCore.Functions.Animation.new(iPed, iType, iText, iDuration, iDict, iAnim, iFlag)
     local self = {}
 
     self.ped = iPed;
@@ -142,7 +142,7 @@ function QBCore.Functions.Animation:new(iPed, iType, iText, iDuration, iDict, iA
     return setmetatable(self, has)
 end
 
-function QBCore.Functions.Animation:start(Task)
+function QBCore.Functions.Animation.start(Task)
     if has.active then return end
 
     has.active = true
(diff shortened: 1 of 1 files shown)


client/functions.lua
```diff
@@ -111,14 +112,86 @@ function QBCore.Functions.GetBlip(id)
...
+        local timeout = false
+        SetTimeout(60000, function () timeout = false end)
+		
+        while not HasAnimDictLoaded(animDict) and not timeout do
+            Wait(0)
+        end
+    end
+end
+
+function QBCore.Functions.Animation:new(iPed, iType, iText, iDuration, iDict, iAnim, iFlag)
```
Mojito-Fivem, line 128: This will break compatibility with other ScRT (javascript and c#) so maybe we shouldn't use metatables and this syntax for functions. Not to mention this doesn't fit with the style and conventions qbcore is already using

client/functions.lua
```diff
@@ -111,14 +112,86 @@ function QBCore.Functions.GetBlip(id)
...
+    local this = {}
```
Mojito-Fivem, line 129: If you absolutely have to use metatables for this, the convention is to call this `self` and not `this`

Mojito-Fivem (changes requested):

Mojito-Fivem:
You're still using the colons for the functions and metatables, this will not work outside of lua

darknight2590:
@Mojito-Fivem Let me know if you have any other comments
thank you

darknight2590:
If you suggest a good form of the phrase, let me know

Mojito-Fivem:
> @Mojito-Fivem Let me know if you have any other comments thank you

It's not this specifically I have a problem with, it's your use here of this pseudo-object oriented metatable system you're trying to implement. It doesn't match the rest of qbcore and isn't compatible across different ScRTs

darknight2590:
When I get back from work, I will try to restructure the issue

PR #393, opened 2021-12-25
make IsWearingHandshoes in core

darknight2590:
**Describe Pull request**

make IsWearingHandshoes in core


Mojito-Fivem:
Just a suggestion, maybe we should call this 'IsWearingGloves' rather than continuing to use a function name which comes from a bad translation

PR #394, opened 2021-12-25
Add GetStreetLabel to core

darknight2590:
**Describe Pull request**

Add GetStreetLabel to core

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -362,6 +362,16 @@ function QBCore.Functions.GetBoneDistance(entity, Type, Bone)
     return #(boneCoords - playerCoords)
 end
 
+QBCore.Functions.GetStreetLabel = function()
+	local s1, s2 = GetStreetNameAtCoord(GetEntityCoords(PlayerPedId()).x, GetEntityCoords(PlayerPedId()).y, GetEntityCoords(PlayerPedId()).z)
+    local street1 = GetStreetNameFromHashKey(s1)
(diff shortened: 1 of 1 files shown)


xenfovn:
i think get form coords is good

client/functions.lua
```diff
@@ -362,6 +362,16 @@ function QBCore.Functions.GetBoneDistance(entity, Type, Bone)
     return #(boneCoords - playerCoords)
 end
 
+QBCore.Functions.GetStreetLabel = function()
+	local s1, s2 = GetStreetNameAtCoord(GetEntityCoords(PlayerPedId()).x, GetEntityCoords(PlayerPedId()).y, GetEntityCoords(PlayerPedId()).z)
```
Mojito-Fivem, line 366: Invoking the native this many times is probably not great

client/functions.lua
```diff
@@ -362,6 +362,16 @@ function QBCore.Functions.GetBoneDistance(entity, Type, Bone)
     return #(boneCoords - playerCoords)
 end
 
+QBCore.Functions.GetStreetLabel = function()
```
Mojito-Fivem, line 365: I think having the options to have a coords parameter would make this more versatile and you can default to the player's coords if it is not specified

Mojito-Fivem (changes requested):

evanillaa (approved):

PR #407, opened 2021-12-31
Couple of changes in cl_functions

JajarGaming:
**Describe Pull request**
I added a couple of changes to the blips and drawtext functions. I added in QBCore.Functions.RotationToDirection and function QBcore.Functions.ScreenRelToWorld because this has helped me in my own server and could help others with several scripts.

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? **No, haven't had the time yet.**
- Does your code fit the style guidelines? **Yes**
- Does your PR fit the contribution guidelines? **Yes**

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -78,7 +78,7 @@ function QBCore.Functions.CreateBlip(id, data)
     if data.route then SetBlipRoute(blip, data.route) end
     if data.friendly then SetBlipAsFriendly(blip, data.friendly) end
     if data.routecolor then SetBlipRouteColour(blip, data.routecolor) end
-    if data.scale then SetBlipScale(blip, data.scale) else SetBlipScale(blip, 0.8) end
+    if data.scale then SetBlipScale(blip, data.scale) else SetBlipScale(blip, 0.7) end
 
     BeginTextCommandSetBlipName("STRING")
     AddTextComponentString(data.name)
@@ -96,13 +96,7 @@ end
 function QBCore.Functions.HideBlip(id, toggle)
     local blip = Blips[id]
     if not blip then return end
-    if toggle then 
-        SetBlipAlpha(blip.blip, 0)
-        SetBlipHiddenOnLegend(blip.blip, true)
-    else
-        SetBlipAlpha(blip.blip, 255)
-        SetBlipHiddenOnLegend(blip.blip, false)
-    end
+    if toggle then SetBlipAlpha(blip, 0) else SetBlipAlpha(blip, 255) end
 end
 
 function QBCore.Functions.GetBlip(id)
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -32,35 +32,100 @@ end
 
 -- Utility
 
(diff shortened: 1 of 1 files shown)


client/functions.lua
```diff
@@ -32,35 +32,100 @@ end
 
 -- Utility
 
-function QBCore.Functions.DrawText(x, y, width, height, scale, r, g, b, a, text)
-    -- Use local function instead
+function drawTxt(x,y ,width,height,scale, text, r,g,b,a)
```
Mojito-Fivem, line 35: This function isn't `local` you've just redeclared it

client/functions.lua
```diff
@@ -32,35 +32,100 @@ end
...
+    
+  local z = math.rad(rotation.z);
+  local x = math.rad(rotation.x);
+  local num = math.abs(math.cos(x));
+
+  local vector3Direction = vector3(-math.sin(z) * num, math.cos(z) * num, math.sin(x))
+  return vector3Direction
+end
+
+function QBcore.Functions.ScreenRelToWorld(camPos,camRot,coord)
```
Mojito-Fivem, line 88: What's the use case for this, why does it need to be in the QBCore object?

client/functions.lua
```diff
@@ -32,35 +32,100 @@ end
...
+
+      local distance = 1000.0
+      local camForward = QBCore.Functions.RotationToDirection(camRot);
+      local rotUp = camRot + vector3(distance, 0, 0);
+      local rotDown = camRot + vector3(-distance, 0, 0);
+      local rotLeft = camRot + vector3(0, 0, -distance);
+      local rotRight = camRot + vector3(0, 0, distance);
+
+      local camRight = QBCore.Functions.RotationToDirection(rotRight) - QBCore.Functions.RotationToDirection(rotLeft);
+      local camUp = TQBCore.Functions.RotationToDirection(rotUp) - QBCore.Functions.RotationToDirection(rotDown);
```
Mojito-Fivem, line 98: `TQBCore` typo

client/functions.lua
```diff
@@ -96,13 +161,7 @@ end
...
     local blip = Blips[id]
     if not blip then return end
-    if toggle then 
-        SetBlipAlpha(blip.blip, 0)
-        SetBlipHiddenOnLegend(blip.blip, true)
-    else
-        SetBlipAlpha(blip.blip, 255)
-        SetBlipHiddenOnLegend(blip.blip, false)
-    end
+    if toggle then SetBlipAlpha(blip, 0) else SetBlipAlpha(blip, 255) end
```
Mojito-Fivem, line 164: This just sets the alpha? At 0 alpha it will still be on the legend in the escape menu so won't be fully hidden

client/functions.lua
Mojito-Fivem: You're also not using this function and declaring it just above the QBCore copy

Mojito-Fivem (changes requested):
This shouldn't all be in a single pull request imo, you should refine the scope of this PR

PR #411, opened 2022-01-01
Split Shared.lua into shared folder with mutiple files - BREAKING CHANGE

IdrisDose:
**Describe Pull request**
I've split the shared file into a shared folder for easier config and maintenance. New file structure:
- shared/main.lua - is where the util functions and starter item config got placed
- shared/items.lua - self explanatory but where the QBShared.Items has been placed
- shared/weapons.lua - same as items but QBShared.Weapons
- shared/jobs.lua - same as above for jobs and the defaultduty config option
- shared/vehicles.lua - same as above for vehicles
- shared/gangs - same but for gans

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? yes, I spawned vehicles and items, used items, inventory worked etc.


Mojito-Fivem:
This looks good to me, I don't see how this is breaking however since the end result is the same unless you're refering to people needing to migrate across files

PR #418, opened 2022-01-02
Revert "edit Functions Blip"

Mojito-Fivem:
Reverts qbcore-framework/qb-core#386


Mojito-Fivem:
Upon further review, I have decided that this code makes 0 sense 🙂

PR #419, opened 2022-01-02
Update weapons.lua

Demo4889:
Adding the new weapons from the Security DLC (Heavy Rifle, EMP Launcher, Fertilizer Can). Heavy Rifle has been tested with Rifle Ammo, and EMP Launcher has been tested with EMP Launcher Ammo. Will put in a PR for qb-weapons to add the EMP Launcher Ammo as well.


evanillaa:
@Demo4889 
Why not add weapon_stungun_mp
Ammo name : AMMO_STUNGUN

See this
https://forum.cfx.re/t/the-contract-infos-build-2545/4792123

Demo4889:
@evanillaa I didn't add the weapon_stungun_mp because when i did, it was exactly the same as the standard stun gun... so why add it?

evanillaa:
Thanks, I will add it tomorrow

PR #424, opened 2022-01-03
Update vehicles.lua

Demo4889:
Adding brand to cycles. All brands are 'Unknown' as there is no actual brand for them.

This is a fix for when people want to add "Cycles" to their shops.

If the brand doesn't exist, it will throw an error in the vehicleshop for a nil value on getVehBrand()

### DIFF
--- a/shared/vehicles.lua
+++ b/shared/vehicles.lua
@@ -320,6 +320,7 @@ QBShared.Vehicles = {
 	--- Cycles
 	['bmx'] = {
 		['name'] = 'BMX',
+        ['brand'] = 'Unknown',
 		['model'] = 'bmx',
 		['price'] = 160,
 		['category'] = 'cycles',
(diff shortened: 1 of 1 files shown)


Demo4889:
Either this PR, or can modify the coding in the vehicleshop for the function... whichever you guys prefer...

`local function getVehBrand()
    local brand = QBCore.Shared.Vehicles[Config.Shops[getShopInsideOf()]["ShowroomVehicles"][ClosestVehicle].chosenVehicle]["brand"]

    if brand then
        return QBCore.Shared.Vehicles[Config.Shops[getShopInsideOf()]["ShowroomVehicles"][ClosestVehicle].chosenVehicle]["brand"]
    else
        return "Unknown"
    end
end`

evanillaa (approved):

JajarGaming (approved):
Looks good to me

FjamZoo:
Cycles already have brands now this one can be closed

PR #429, opened 2022-01-05
addition function IsInsideArea

darknight2590:
**Describe Pull request**

Add more functions inside qbcore

- Does your code fit the style guidelines? [yes
- Does your PR fit the contribution guidelines? [yes/

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -324,6 +324,16 @@ function QBCore.Functions.GetClosestObject(coords)
     return closestObject, closestDistance
 end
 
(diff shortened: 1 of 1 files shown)


YeahItsHayden (approved):
Very basic and not really needed. It'd be faster to just add this into your scripts (using localised functions would be faster than calling directly to the core).

Nonetheless, it makes it easier for those who are lazy, as such, I approve

darknight2590:
I am one of those lazy people you talked about

Mojito-Fivem (changes requested):
Too simple, no comments, template not filled in.

PR #439, opened 2022-01-09
GetStreet And Zone shared functions

darknight2590:
**Describe Pull request**

GetStreet And Zone shared functions

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -96,6 +96,16 @@ function QBCore.Functions.LoadModel(ModelName)
 	end
 end
 
+function QBCore.Functions.GetStreetAndZone()
+    local pos = GetEntityCoords(PlayerPedId(), true)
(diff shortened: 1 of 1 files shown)


client/functions.lua
```diff
@@ -96,6 +96,16 @@ function QBCore.Functions.LoadModel(ModelName)
...
 end
 
+function QBCore.Functions.GetStreetAndZone()
+    local pos = GetEntityCoords(PlayerPedId(), true)
+	local s1, s2 = GetStreetNameAtCoord(pos.x, pos.y, pos.z)
+    local street1 = GetStreetNameFromHashKey(s1)
+    local street2 = GetStreetNameFromHashKey(s2)
+    local zone = GetLabelText(GetNameOfZone(pos.x, pos.y, pos.z))
+    local street = street1 .. ", " .. zone
+    return street
```
IdrisDose, line 106: Could probably `return street1, street2, zone` instead of a concatenated string might be good.

PR #441, opened 2022-01-09
make PromiseTimeout in shared functions

darknight2590:
**Describe Pull request**

make PromiseTimeout in shared functions

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -29,6 +29,16 @@ function QBCore.Functions.HasItem(item)
     return Citizen.Await(p)
 end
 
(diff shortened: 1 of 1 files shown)


darknight2590:
using 

> local out = PromiseTimeout(1000)

Mojito-Fivem:
Just use Wait, this is unnecessary

TasoOneAsia:
Have to agree with Mojito in this instance. API seems a little weird and convoluted and I can't see a user use case that wouldn't devolve into spaghetti with this approach.

PR #442, opened 2022-01-10
security: Deprecated the AddItem, Remove Item Set Meta Data events

IdrisDose:
General security standards depict never accepting client input as truth. This will close major exploits where events can be directly called and manipulate player data.

### DIFF
--- a/server/events.lua
+++ b/server/events.lua
@@ -139,22 +139,10 @@ RegisterNetEvent('QBCore:UpdatePlayer', function()
 end)
 
 RegisterNetEvent('QBCore:Server:SetMetaData', function(meta, data)
-    -- local src = source
-    -- local Player = QBCore.Functions.GetPlayer(src)
-    -- if meta == 'hunger' or meta == 'thirst' then
-        -- if data > 100 then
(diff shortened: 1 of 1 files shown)


YeahItsHayden (approved):
Definitely needed

izMystic (approved):
expoit boig

darknight2590:
There is a lot in the scripts that needs to be modified to suit your modification

Mojito-Fivem:
Why is this code commented out when were using version control 🤨

IdrisDose:
> Why is this code commented out when were using version control 🤨

Commented out just for reference in case some dingus' want to use old code they copy it and make a new function somewhere.

> There is a lot in the scripts that needs to be modified to suit your modification
Yes, We're not merging it yet already spoken to Kakarot about it and put a message out to the dev team to keep note of where the implementations are within the qb resources do to do their best to move the logic to the server-side.

JajarGaming (approved):

JajarGaming (approved):

GhzGarage:
#559

PR #457, opened 2022-01-10
Spanish locale

CherryozZ:


CherryozZ:
I'm a native spanish speaker just in case :P

PR #462, opened 2022-01-11
fix(client/functions): Return right result when with hasitem callback

BerkieBb:
**Describe Pull request**
This fixes the HasItem function always returning false


IdrisDose (approved):
This would indeed call the correct result of the promise.

evanillaa (approved):
Now its use is multi-items
thank you

PR #479, opened 2022-01-13
Callback RemoveMoney

darknight2590:
**Describe Pull request**
It is clear from the title RemoveMoney Callback 

- Does your code fit the style guidelines? [yes/
- Does your PR fit the contribution guidelines? [yes/

### DIFF
--- a/server/events.lua
+++ b/server/events.lua
@@ -188,6 +188,16 @@ RegisterNetEvent('QBCore:Server:AddItem', function(itemName, amount, slot, info)
     Player.Functions.AddItem(itemName, amount, slot, info)
 end)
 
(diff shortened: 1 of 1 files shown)


Mojito-Fivem:
template not filled in, no comments, no explanation. Why is this even needed (hint: it's not because QBCore.Functions.RemoveMoney already returns a boolean)

IdrisDose:
Denied. As Mojito said It's not needed since the server-side function already returns a boolean value. Should not be introducing events exposed to the client to directly manipulate player data.

darknight2590:
It's close  , I'm using it in my core

PR #488, opened 2022-01-15
feat: add helicopters into Spawn vehicles menu

jacanales:
**Describe Pull request**
Add Helicopters into vehicle list, so they can be spawned from the Admin Menu.

**Questions (please complete the following information):**
- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? Yes. Adding screenshopts at the end of the description.
- Does your code fit the style guidelines? yes
- Does your PR fit the contribution guidelines? yes

**Show helicoptes category in vehicles menu**
menu-vehicles

**Show helicopters list after enter the helicopters category**
menu-helicopters

PR #492, opened 2022-01-16
Update vehicles.lua

Demo4889:
Adding boats, helis and planes to the vehicles file for future use.


Mojito-Fivem:
Duplicate of #488

Demo4889:
> Duplicate of #488

This one adds more than just the helicopters as the other one only has helicopters

JajarGaming:
> Adding boats, helis and planes to the vehicles file for future use.

Template isn't added and filled out

PR #514, opened 2022-01-27
[Update] Missing Vehicle Properties

ItsANoBrainer:
Added proper support for missing or unfinished vehicle property saving/getting:

- neons enabled/disabled
- tire health
- tire burst state
- tire burst completely
- oil level
- wheel size
- wheel width
- headlight color
- interior color
- modkit 17, 19, 21, 47, 49
- livery roof
- tankHealth

- Set LegacyFuel on setting fuel level

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -61,7 +61,9 @@ RegisterNetEvent('QBCore:Command:SpawnVehicle', function(vehName)
     end
     local vehicle = CreateVehicle(hash, GetEntityCoords(ped), GetEntityHeading(ped), true, false)
     TaskWarpPedIntoVehicle(ped, vehicle, -1)
+    exports['LegacyFuel']:SetFuel(vehicle, 100)
     SetModelAsNoLongerNeeded(vehicle)
+    
     TriggerEvent("vehiclekeys:client:SetOwner", QBCore.Functions.GetPlate(vehicle))
 end)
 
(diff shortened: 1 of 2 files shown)


JajarGaming (approved):
LGTM

Mojito-Fivem:
Is it a great idea to have LegacyFuel be a dependency of qb-core? Maybe in other resources sure but I don't think it's a great idea to restrict versatility of the base framework in my opinion

ItsANoBrainer:
> 

I mean, LegacyFuel is apart of the qb-core framework project repo and tx-admin recipe. At this point isn't it already a dependency?

JnKTechstuff:
> Is it a great idea to have LegacyFuel be a dependency of qb-core? Maybe in other resources sure but I don't think it's a great idea to restrict versatility of the base framework in my opinion

I would honestly just have a Config option for them to use their fuel script export of their choice. Like how some escrow scripts have their stuff setup.

darknight2590:
Thank you for the beautiful edits
We are currently using LegacyFuel 
I don't know why the words are useful in responses
The project originally used this resource LegacyFuel 
But there is a saying that says I was created to object

darknight2590 (approved):

IdrisDose:
Hey, I like the edit, however, Mojito is correct can you remove the LegacyFuel export from the core please :)

evanillaa:
@ItsANoBrainer

ItsANoBrainer:
Done

@IdrisDose @evanillaa

BerkieBb:
Instead of removing it, you could use the native SetVehicleFuelLevel

ItsANoBrainer:
> Instead of removing it, you could use the native SetVehicleFuelLevel

Good idea, for some reason I thought that was already done, added.

PR #516, opened 2022-02-01
feat(qb-core/player): Added server-side event onJobUpdate

chainiDV:
Added server-side event onJobUpdate so scripts can listen on server scripts if any player has changed their job. It's safer than redirecting client event to server again (never trust the client).


JajarGaming (approved):

Mojito-Fivem (approved):
Makes sense

PR #536, opened 2022-02-18
[dev branch] Some additions

BerkieBb:
**Describe Pull request**
- Lots of functions changed to have less indents
- Changed lots of unnecessary `local src = source` to just source
- Changed some variables to camelCase
- Small bits of optimisation
- QBCore.Debug now accepts more than a table
- Added the OnPermissionUpdate event to the RemovePermission function
- Removed unused LoggedIn PlayerData
- changed all variable assigned functions to use function syntax
- Added `Player.Functions.Logout` function
- Enabled globbing for the shared files to make it easier to add new shared files
- Added a new permission system which utilizes ace permissions
- Added whitelist functionality
- Added the functionality to open and close the server (pretty much the check to close the server was missing)
- Made connectqueue optional instead of a dependency
- Removed progressbar as a dependency and instead make a very obvious error when it's not started when the function is used
- Changed some functions to be more 'flexible'
- Fixed issue with people having removed gangs and/or jobs - @ItsANoBrainer 
You need this in your server.cfg for the new permission system to work with ace permissions:
`add_ace resource.qb-core command allow`


evanillaa:
oh nice thank you for big edit

fxmanifest.lua
```diff
@@ -2,18 +2,13 @@ fx_version 'cerulean'
...
 	'shared/locale.lua',
 	'locale/en.lua', -- replace with desired language
 	'config.lua',
-	'shared/main.lua',
-	'shared/items.lua',
-	'shared/jobs.lua',
-	'shared/vehicles.lua',
-	'shared/gangs.lua',
-	'shared/weapons.lua'
+	'shared/*.lua'
```
IdrisDose, line 11: please add shared/main.lua first and then globbing (just due to the fact that main is QBShared = {} the rest are QBShared = QBShared or {} )

server/exports.lua
```diff
@@ -145,9 +145,7 @@ exports('AddGangs', AddGangs)
 
 local function GetCoreVersion(InvokingResource)
     local resourceVersion = GetResourceMetadata(GetCurrentResourceName(), 'version')
-    if InvokingResource ~= nil then
-        print( ("%s called qbcore version check: %s"):format(InvokingResource, resourceVersion) )
-    end
+    print(("%s called qbcore version check: %s"):format(InvokingResource or 'Unknown Resource', resourceVersion))
```
IdrisDose, line 148: Left the if statement in specifically to nil check so they dont have to pass in the invoking resource, should've added a comment my bad.

server/events.lua
```diff
@@ -101,7 +97,7 @@ end)
 
 RegisterNetEvent('QBCore:server:OpenServer', function()
     local src = source
-    if QBCore.Functions.HasPermission(src, 'admin') or QBCore.Functions.HasPermission(src, 'god') then
+    if QBCore.Functions.HasPermission(src, 'admin') then
```
IdrisDose, line 100: Requires god as well unless we restructure the permission system, god is considered higher than admin.

server/events.lua
```diff
@@ -90,8 +86,8 @@ AddEventHandler('playerConnecting', OnPlayerConnecting)
 
 RegisterNetEvent('QBCore:server:CloseServer', function(reason)
     local src = source
-    if QBCore.Functions.HasPermission(src, 'admin') or QBCore.Functions.HasPermission(src, 'god') then
-        local reason = reason or 'No reason specified'
+    if QBCore.Functions.HasPermission(src, 'admin') then
```
IdrisDose, line 89: Requires god as well unless we restructure the permission system, god is considered higher than admin.

IdrisDose (changes requested):
Just some edits needed, Please review them.

fxmanifest.lua
BerkieBb: If you look at shared/main.lua I changed it to `QBShared = QBShared or {}` so it can load like the rest, the functions don't depend on each other in the shared so the loading order doesn't matter that way

server/events.lua
BerkieBb: The HasPermission function allows it already even if you check for another permission that isn't god

server/events.lua
BerkieBb: The HasPermission function allows it already even if you check for another permission that isn't god

server/exports.lua
BerkieBb: My bad, I overlooked this, will change this one back

BerkieBb:
> Would also prefer not to use `-=` and `+=` looks cleaner yes, but preference stick to traditional method.

Well, if lua 5.4 is enabled in the fxmanifest, why not use it? 😋

evanillaa:
I did not encounter any problems after the test

evanillaa (approved):

evanillaa:
@BerkieBb conflicts

BerkieBb:
@IdrisDose fixed conflicts and reviews replied to, please re-review

IllInuz:
Tested this, can confirm working without issues.

BerkieBb:
`add_ace resource.qb-core command allow`
Add this to your server.cfg for the permissions system to add the aces

iLostMyDriversLicense (approved):
LGTM!

KingRyuShin:
@BerkieBb Since I added thoses change on my server, I am no longer able to any use admin command even tho in the config UseOldPermissionSystem is set to true and in my SQL Database my permission are set.

image

how do you add yourself admin then ?

BerkieBb:
> @BerkieBb Since I added thoses change on my server, I am no longer able to any use admin command even tho in the config UseOldPermissionSystem is set to true and in my SQL Database my permission are set.
> 
> image
> 
> how do you add yourself admin then with the new system since the old one doesn't seem to work for me?

The old still works, I tested that thoroughly so you must've done something wrong there, but to give yourself god permission you have to give yourself group.admin in the cfg using principals, after that you can use the addpermission and removepermission command (once you restarted the server to get the permission) to add other people amd it'll save to the database in PlayerData, the property is PlayerData.permission

KingRyuShin:
> > @BerkieBb Since I added thoses change on my server, I am no longer able to any use admin command even tho in the config UseOldPermissionSystem is set to true and in my SQL Database my permission are set.
> > image
> > how do you add yourself admin then with the new system since the old one doesn't seem to work for me?
> 
> The old still works, I tested that thoroughly so you must've done something wrong there, but to give yourself god permission you have to give yourself group.admin in the cfg using principals, after that you can use the addpermission and removepermission command (once you restarted the server to get the permission) to add other people amd it'll save to the database in PlayerData, the property is PlayerData.permission

I've already given myself permission with the group admin
image

I've try using /addpermission 1 god and it does nothing

BerkieBb:
Fixed the issue @KingRyuShin

darknight2590 (approved):

server/exports.lua
```diff
@@ -126,7 +126,7 @@ local function AddGangs(gangs)
             break
         end
 
-        if QBCore.Shared.Gangs[key] ~= nil then
+        if QBCore.Shared.Gangs[key] then
             message = "gang_exists"
             shouldContinue = false
             errorItem = items[key]
```
ItsANoBrainer, line 132:
```suggestion
            errorItem = gangs[key]
```

client/functions.lua
```diff
@@ -865,18 +857,22 @@ function QBCore.Functions.SetVehicleProperties(vehicle, props)
...
 
-function QBCore.Functions.StartParticleAtCoord(Dict, ptName, looped, coords, rot, scale, alpha, color, duration)
-    QBCore.Functions.LoadParticleDictionary(Dict)
-    UseParticleFxAssetNextCall(Dict)
-    SetPtfxAssetNextCall(Dict)
+function QBCore.Functions.StartParticleAtCoord(dict, ptName, looped, coords, rot, scale, alpha, color, duration)
+    if coords then
+        coords = type(coords) == 'table' and vec3(coords.x, coords.y, coords.z) or coords
+    else
+        coords = GetEntityCoords(ped)
```
ItsANoBrainer, line 871:
```suggestion
        coords = GetEntityCoords(PlayerPedId())
```
Line 405 needs to be fixed with this as well.

fxmanifest.lua
```diff
@@ -46,8 +40,7 @@ files {
 
 dependencies {
 	'oxmysql',
-	'progressbar',
-	'connectqueue'
+	'progressbar'
```
ItsANoBrainer, line 43:
```suggestion

```
Can we remove this from the fxmanifest?

server/exports.lua
ItsANoBrainer, line 132:
```suggestion
            errorItem = gangs[key]
```

Line 125 should also be changed to this.

server/functions.lua
```diff
@@ -173,40 +164,37 @@ function QBCore.Functions.GetEntitiesInBucket(bucket --[[int]])
...
-                    if account < payment then -- Checks if company has enough money to pay society
-                        TriggerClientEvent('QBCore:Notify', Player.PlayerData.source, Lang:t('error.company_too_poor'), 'error')
+function PaycheckInterval()
+    if next(QBCore.Players) then
+        for _, Player in pairs(QBCore.Players) do
+            if Player then
+                local payment = Player.PlayerData.job.payment
+                if Player.PlayerData.job and payment > 0 and (QBShared.Jobs[Player.PlayerData.job.name].offDutyPay or Player.PlayerData.job.onduty) then
+                    if QBCore.Config.Money.PayCheckSociety then
+                        local account = exports['qb-bossmenu']:GetAccount(Player.PlayerData.job.name)
```
ItsANoBrainer, line 176:
```suggestion
                        local account = exports['qb-management']:GetAccount(Player.PlayerData.job.name)
```
When are we switching this over?

server/player.lua
```diff
@@ -523,37 +502,35 @@ local playertables = { -- Add tables as needed
...
 			if result then
-				TriggerEvent('qb-log:server:CreateLog', 'joinleave', 'Character Deleted', 'red', '**' .. GetPlayerName(src) .. '** ' .. license .. ' deleted **' .. citizenid .. '**..')
+				TriggerEvent('qb-log:server:CreateLog', 'joinleave', 'Character Deleted', 'red', '**' .. GetPlayerName(source) .. '** ' .. license .. ' deleted **' .. citizenid .. '**..')
             end
 		end)
     else
-        DropPlayer(src, 'You Have Been Kicked For Exploitation')
-        TriggerEvent('qb-log:server:CreateLog', 'anticheat', 'Anti-Cheat', 'white', GetPlayerName(src) .. ' Has Been Dropped For Character Deletion Exploit', false)
+        DropPlayer(source, 'You Have Been Kicked For Exploitation')
+        TriggerEvent('qb-log:server:CreateLog', 'anticheat', 'Anti-Cheat', 'white', GetPlayerName(source) .. ' Has Been Dropped For Character Deletion Exploit', false)
```
ItsANoBrainer, line 524:
```suggestion
        TriggerEvent('qb-log:server:CreateLog', 'anticheat', 'Anti-Cheat', 'white', GetPlayerName(source) .. ' Has Been Dropped For Character Deletion Exploit', true)
```

ItsANoBrainer (changes requested):

fxmanifest.lua
BerkieBb: Not yet as the progressbar function is still needed as a dependency, people are going to get errors when they use it without that so when it's a dependency it will show they need it aswell before they can start the core

server/functions.lua
BerkieBb: Not yet as it's not in the recipe as of now, qb-bossmenu still is

server/player.lua
BerkieBb: You don't want to be tagged here as this log already goes into it's own channel for logs already, so it's already exceptional

darknight2590:
add new functions

> function QBCore.Functions.GetVehicleInDirection()
> 	local playerPed    = PlayerPedId()
> 	local playerCoords = GetEntityCoords(playerPed)
> 	local inDirection  = GetOffsetFromEntityInWorldCoords(playerPed, 0.0, 5.0, 0.0)
> 	local rayHandle    = StartShapeTestRay(playerCoords, inDirection, 10, playerPed, 0)
> 	local numRayHandle, hit, endCoords, surfaceNormal, entityHit = GetShapeTestResult(rayHandle)
> 
> 	if hit == 1 and GetEntityType(entityHit) == 2 then
> 		return entityHit
> 	end
> 
> 	return nil
> end

ItsANoBrainer (approved):

Issue #530, opened 2022-02-13
[NEW FEATURE] Modular Shared Tables


IdrisDose:
I've added the ability to dynamically add new items, jobs, and gangs to the shared without having to edit the shared files.

I've added the following functions: 
|Function or Event | Scope | Description | returns |
|--|--|--|--|
| exports['qb-core']:AddItem(itemName, item) |  Server | Accepts an item name and item object - will add new entry to QBCore.Shared.Items. | boolean, string |
| exports['qb-core']:AddJob(jobName, job) |  Server | Accepts an job name and job object - will add new entry to QBCore.Shared.Jobs.| boolean, string |
| exports['qb-core']:AddGang(gangName, gang) |  Server | Accepts an gang name and gang object - will add new entry to QBCore.Shared.Gangs.| boolean, string |
| exports['qb-core']:AddItems(items) |  Server | Accepts a table or array - will loop through the table and add new entries to QBCore.Shared.Items | boolean, string, errorItem |
| exports['qb-core']:AddJobs(jobs) |  Server | Accepts a table or array - will loop through the table and add new entries to QBCore.Shared.Jobs | boolean, string, errorItem |
| exports['qb-core']:AddGangs(gangs) |  Server | Accepts a table or array - will loop through the table and add new entries to QBCore.Shared.Gangs | boolean, string, errorItem |
| QBCore.Functions.AddItem(itemName, item) |  Server | Accepts an item name and item object - will add new entry to QBCore.Shared.Items | boolean, string |
| QBCore.Functions.AddJob(jobName, Job) |  Server | Accepts an job name and job object - will add new entry to QBCore.Shared.Jobs | boolean, string |
| QBCore.Functions.AddGang(gangName, gang) |  Server | Accepts an gang name and gang object - will add new entry to QBCore.Shared.Gangs | boolean, string |
| QBCore.Functions.AddItems(items) |  Server | Accepts a table or array - will loop through the table and add new entries to QBCore.Shared.Items | boolean, string, errorItem |
| QBCore.Functions.AddJobs(jobs) |  Server | Accepts a table or array - will loop through the table and add new entries to QBCore.Shared.Jobs | boolean, string, errorItem |
| QBCore.Functions.AddGangs(gangs) |  Server | Accepts a table or array - will loop through the table and add new entries to QBCore.Shared.Gangs| boolean, string, errorItem |
[...]

Issue #527, opened 2022-02-13
[NEW FEATURE] DrawText in core


IdrisDose:
I've added the drawtext feature from [qb-drawtext](https://github.com/idrisdose/qb-drawtext).
# qb-drawtext

A NUI DrawText with a few different config options. Please make sure you handle the showing and hiding as if done incorrectly it will just repeatadly send nui messages.

## How to use

This can be triggered from the server or client. A more advanced example is posted below.
|Function or Event | Description |
|--|--|
| exports['qb-core']:DrawText('message','position')| This is a client export that will draw a message at the specified position (listed below) |
| exports['qb-core']:ChangeText('message','position')| This is a client export that will change the currently displayed message at the specified position (listed below) |
| exports['qb-core']:HideText() | This will hide the text display |
| exports['qb-core']:KeyPressed() | This is useful if you want to change the background and hide the text on keypress (if not handled correctly users will have to renter the zone to display) |
| TriggerClientEvent('qb-core:client:DrawText', source, message, position) | The same as the export but as an event |
| TriggerClientEvent('qb-core:client:ChangeText', source, message, position) | The same as the export but as an event |
| TriggerClientEvent('qb-core:client:HideText', source) | The same as the export but as an event |
| TriggerClientEvent('qb-core:client:KeyPressed', source) | The same as the export but as an event |

Positions accepted are:

* left
* right
* top

PR #554, opened 2022-03-01
Dev branch merge into Main

IdrisDose:
**Describe Pull request**
This PR implements some new functionality as well as some fixes.

**Features**
- DrawText is now implemented into Core meaning you can display text on either top, left or right. See Issue #527
- Dynamic Shared Tables - Added the ability to dynamically add gangs, jobs and items to the core at resource startup without having to create a new shared entry. See issue #530 

**Fixes**
- Added missing vehicle properties.


client/drawtext.lua
```diff
@@ -0,0 +1,60 @@
...
+        action = 'CHANGE_TEXT',
+        data = {
+            text = text,
+            position = position
+        }
+    })
+end
+
+local function keyPressed()
+    CreateThread(function() -- Not sure if a thread is needed but why not eh?
```
deandum, line 32: Doesn't this start a thread for each key pressed? I don't think this is needed. I tested it locally and it works fine without it.

BerkieBb: That's not how that works

BerkieBb: But yeah it's unnecessary

JonasDev99:
On line 600 there is a += which afaik there is no += in lua.

```
function QBCore.Player.GetTotalWeight(items)
    local weight = 0
    if not items then return 0 end
    for _, item in pairs(items) do
        weight += item.weight * item.amount
    end
    return tonumber(weight)
end
```

BerkieBb:
> On line 600 there is a += which afaik there is no += in lua.
> 
> ```
> function QBCore.Player.GetTotalWeight(items)
>     local weight = 0
>     if not items then return 0 end
>     for _, item in pairs(items) do
>         weight += item.weight * item.amount
>     end
>     return tonumber(weight)
> end
> ```

There is in cfxlua
https://github.com/citizenfx/lua/tree/luaglm-dev/cfx
Cfxlua has a lot which a lot of people don't knoe about

JonasDev99:
Oh, didn't know, nice 👍

PR #570, opened 2022-03-19
Update ar.lua

elvodkaa:


evanillaa (approved):

evanillaa:
شكرا لك

evanillaa:
The translation here is better than before

PR #571, opened 2022-03-20
Prevent people using /me to troll

Scullyy:
**Describe Pull request**
This prevents people using that /me troll with the huge font size and different colors.

### DIFF
--- a/server/commands.lua
+++ b/server/commands.lua
@@ -240,6 +240,10 @@ QBCore.Commands.Add('me', 'Show local message', {{name = 'message', help = 'Mess
     local pCoords = GetEntityCoords(ped)
     local msg = table.concat(args, ' ')
     if msg == '' then return end
+    if string.match(msg, "<") then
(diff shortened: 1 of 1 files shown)


zfbx:
input should actually just be fully sanitized/escaped. If I'm not mistaken there are other exploits to inject code without explicitly `<` and also for people who like to use them this would trigger any message that types `<3` as well. at least just replace them all with `&lt;`. just my 2 cents

Scullyy:
> input should actually just be fully sanitized/escaped. If I'm not mistaken there are other exploits to inject code without explicitly `<` and also for people who like to use them this would trigger any message that types `<3` as well. at least just replace them all with `&lt;`. just my 2 cents

/me is used to roleplay out an action so you shouldn’t be using it to do <3, this is just to prevent people from doing 

/ME <FONT SIZE = '1000'> TROLL MESSAGE

PR #595, opened 2022-03-28
feat: add global switch function to qb-core

Millusion:
Instead of nesting if/else or having to compare multiple values to the same variable it would be awesome if lua had the switch/case statement. 

By default it hasn't so i thought it might be a good addition to the core.

usage:
local a =1 (or however you want to set your argument to the switch)

switch(a, { 
    [1] = function()    -- for case 1
        print "Case 1."
    end,
    [2] = function()    -- for case 2
        print "Case 2."
    end,
    [3] = function()    -- for case 3
        print "Case 3."
    end
   ['default'] = function() --for default
        print "Default"
   end
})

### DIFF
--- a/shared/main.lua
+++ b/shared/main.lua
@@ -70,6 +70,13 @@ function QBShared.SetDefaultVehicleExtras(vehicle, config)
     end
 end
 
+_G.switch = function(param, case_table)
+    local case = case_table[param]
+    if case then return case() end
+    local def = case_table['default']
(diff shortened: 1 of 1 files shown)


Mojito-Fivem:
This seems so pointless and probably less performant than if chaining

IdrisDose:
While I can see the use case for those who like to use switches, I don't think we want to be including this into the core.

PR #611, opened 2022-03-31
:bug: Remove compound assignment operator

jnccloud:
**Describe Pull request**
Compound assignment operators are not valid in Lua 5.4 without applying a patch (http://lua-users.org/wiki/LuaPowerPatches). This will convert this calculation back to the original method which should not throw any errors on usage.

image
image

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [yes/no] No - but have tested the math and function in a standard Lua 5.4.4 environment.
- Does your code fit the style guidelines? [yes/no] yes
- Does your PR fit the contribution guidelines? [yes/no] yes

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -578,7 +578,7 @@ function QBCore.Player.GetTotalWeight(items)
     local weight = 0
     if not items then return 0 end
     for _, item in pairs(items) do
-        weight += item.weight * item.amount
+        weight = weight + (item.weight * item.amount)
     end
     return tonumber(weight)
 end


BerkieBb:
Read this
https://github.com/citizenfx/lua/tree/luaglm-dev/cfx

jnccloud:
> Read this https://github.com/citizenfx/lua/tree/luaglm-dev/cfx (this is the repo for cfxlua, the lua runtime used with FiveM)

Wow, OK, learn something new every day. Thanks!

PR #613, opened 2022-04-02
feat: Added ChatEvent handler to server/events.lua

IdrisDose:
I've re-added the ChatEvent handler to cancel event if the command does not exist.

### DIFF
--- a/server/events.lua
+++ b/server/events.lua
@@ -1,14 +1,10 @@
 -- Event Handler
 
 AddEventHandler('chatMessage', function(source, _, message)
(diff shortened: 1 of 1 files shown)


MatthewRorke:
Thank you, I can now try updating my production server to the latest version of core <3

MadManDanCat:
There is a bug with this where it still brings the chat up it just doesnt show the command

Andyauk:
> There is a bug with this where it still brings the chat up it just doesn't show the command

I agree.

PR #617, opened 2022-04-04
PlayerData.job.isleo

Lionh34rt:
**Describe Pull request**
Talked to Idris about this briefly, this allows servers to easily make new LEO Jobs like SASP, SAPR and BCSO and allows to easily check if the player is an LEO by using PlayerData.job.isleo.

Feature, not a fix

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [yes/no] (Be honest) Yes
- Does your code fit the style guidelines? [yes/no] Yes
- Does your PR fit the contribution guidelines? [yes/no] Yes


KingRyuShin:
What is this for though ?
Can't we just add a check like any other job like : if (PlayerData.job = "police" or PlayerData.job = "SARP") then ...
It would do the same thing no ?

Lionh34rt:
> What is this for though ? Can't we just add a check like any other job like : if (PlayerData.job = "police" or PlayerData.job = "SARP") then ... It would do the same thing no ?

If jobs and scripts are reworked this way by using .isleo and .ismechanic and .isambulance, server owners can create new jobs for these factions without having to adjust scripts in the future. You would have to adjust every script every time you decide to make a new police department or a new mechanic job and hope you don't forget anything.

That said, it would require every script to be rewritten one time to implement the .isleo variable but afterwards you wouldn't have to touch it again. This would also allow foreign languages to have different names for their police jobs since the locales system show that QBCore wants an easy to setup framework for not only English speaking servers.

Do I think this is a good thing to implement for QBCore Framework?
- Unsure, but I've contacted Idris about it because I'm unsure what direction QBCore wants to head. I know a lot of servers would want to have a Sheriff job, a State Police job etc. in the future and having the possibility to split their jobs so it allows them to split their funds, employees, ...

Kendro1996:
Hey man this is very cool can you give me a idea of what else scripts I need to edit to make this work sorry I'm very new to all this

FjamZoo (approved):
This would create a overall easier user experience as they can skip all the annoying small job checks and just use a variable within the jobs.

ItsANoBrainer (approved):
Big Brain idea, I approve. This beats what I did (having a shared function that all the scripts call).

ItsANoBrainer:
What do we think about changing it from isleo and adding a type entry to each job that can be indexed instead of isleo? That way if in the future people want to add more similar verisons (isambulance, ismechanic) they dont have to edit the player class in qb-core?

Lionh34rt:
> What do we think about changing it from isleo and adding a type entry to each job that can be indexed instead of isleo

Probably a better idea if you want to expand this for all the jobs and not just police.

server/player.lua
```diff
@@ -113,6 +113,7 @@ function QBCore.Player.CheckPlayerData(source, PlayerData)
     PlayerData.job.name = PlayerData.job.name or 'unemployed'
     PlayerData.job.label = PlayerData.job.label or 'Civilian'
     PlayerData.job.payment = PlayerData.job.payment or 10
+    PlayerData.job.isleo = PlayerData.job.isleo or false
```
ItsANoBrainer, line 116:
```suggestion
    PlayerData.job.type = PlayerData.job.type or 'none'
```

ItsANoBrainer, line 116:
```suggestion
    PlayerData.job.type = PlayerData.job.type or 'none'
```

server/player.lua
```diff
@@ -167,6 +168,7 @@ function QBCore.Player.CreatePlayer(PlayerData)
         self.PlayerData.job.name = job
         self.PlayerData.job.label = QBCore.Shared.Jobs[job].label
         self.PlayerData.job.onduty = QBCore.Shared.Jobs[job].defaultDuty
+        self.PlayerData.job.isleo = QBCore.Shared.Jobs[job].isleo or false
```
ItsANoBrainer, line 171:
```suggestion
        self.PlayerData.job.type = QBCore.Shared.Jobs[job].type or 'none'
```

shared/jobs.lua
```diff
@@ -14,6 +14,7 @@ QBShared.Jobs = {
 	},
 	['police'] = {
 		label = 'Law Enforcement',
+        isleo = true,
```
ItsANoBrainer, line 17:
```suggestion
        type = "leo",
```

ItsANoBrainer (changes requested):

Lionh34rt:
@ItsANoBrainer do I have to do anything with these proposed changes or is that for the dev team?

I'm not so familiar with using github

izMystic:
you would commit to your branch with those changes

ItsANoBrainer:
@Lionh34rt On the github comment page here, you should have a "accept changes" or something button by my requested changes.

BerkieBb (approved):

BerkieBb:
Now Idris is left to review this

xenfovn:
good job , If you don't know what it's for, here's a sample :
old
`if PlayerData.job.name == 'police' or PlayerData.job.name == 'ambunce' or PlayerData.job.name == 'mechanic' then`
just use 
`if PlayerData.job.type == "leo" then' then`

ItsANoBrainer (approved):

Kendro1996:
> cant seem to get armoury working can you spot what im doing wrong     

CreateThread(function ()
    Wait(1000)
    while true do
        local sleep = 1000
        if inAmoury and PlayerJob.type == "leo" then
            if onDuty then sleep = 5 end
            if IsControlJustReleased(0, 38) then
                TriggerEvent("qb-police:client:openArmoury")
            end
        else
            sleep = 1000
        end
        Wait(sleep)
    end
end)

PR #649, opened 2022-04-18
server sided vehicle spawning

amoniaclouds:
prevents vehicles from disappearing throughout the gameplay.


TheSpaceGamerV2:
Maybe u should use CREATE_AUTOMOBILE who is a server-side native instead of CreateVehicle who is a RPC?

PR #652, opened 2022-04-20
Add ability to use QBCore:CallCommand to call non-qb commands

IngPleb:
**Describe Pull request**
Added ability to use `QBCore:CallCommand` event to call non-qb commands. This is useful for resources like qb-commandbinding.

- Have you personally loaded this code into an updated qbcore project and checked all its functionality? Yes


sol-fitzy (approved):
Seems like a good idea, code works fine on my half.

PR #655, opened 2022-04-22
Fix give fuel

milotiro95:
Fix for /car admin commands getting fuel not working. Now it does :)


BerkieBb:
This doesn't fix anything, LegacyFuel does the exact same thing, but LegacyFuel also sets an extra variable so it saves how much fuel there is for the resource itself to manage all fuel

BerkieBb:
If you look at the [SetFuel export in LegacyFuel](https://github.com/qbcore-framework/LegacyFuel/blob/master/functions/functions_client.lua#L7) you can see it does the exact same

BerkieBb:
I can't explain that, all I can say is that I've never had this issue

milotiro95:
I have don't test but perhaps it's simple as that 
`SetVehicleFuelLevel(vehicle, 100 + 0.0)`

PR #663, opened 2022-04-25
Better support for 4k screens

JackBro2022:


JackBro2022:
> Great, prefect tested it and works wonderful!

Yo check out pull requests again I made another 4k fix there is a 2nd file from me

PR #669, opened 2022-04-26
Proper optin for staff

ChatDisabled:


ChatDisabled:
Toggling your optin didn't work since you were trying to save a metadata that simply didn't exist. This meant that the command reporttoggle didn't work. Luckely IsOptin has a built in HasPermission function otherwise the whole server would simple see all the staff chats and all the reports.

GhzGarage:
https://github.com/qbcore-framework/qb-core/blob/9f0c613b37aa00d7f2f4b168e22b1af891b699f5/server/player.lua#L47

PR #683, opened 2022-05-04
[Feature] Client Callbacks

ItsANoBrainer:
- Creates support for utilization of Client Callbacks


BerkieBb:
You could cleam up the history a bit 😅

TheiLLeniumStudios:
@BerkieBb we should enable squash commits instead. That way only a single commit per PR goes into main

FjamZoo:
Looks like nobrainer really trynna make it seem like a big pr. 

Looks good ill get it into testing and come back with some results but the code looks good.

PR #684, opened 2022-05-05
[Feature] Client Callbacks

ItsANoBrainer:
Creates support for the utilization of Client Callbacks.

client.lua
```lua
local password = 'hunter2'
QBCore.Functions.CreateClientCallback('example:client:GetClientPassword', function(cb, username)
    print('User requested password: ', username) -- prints 'IdrisSmells'
    cb(password) -- Sends the password back to the server
end)
```

server.lua
```lua
RegisterCommand('requestpassword', function(src)
    QBCore.Functions.TriggerClientCallback('example:client:GetClientPassword', src, function(password)
        print('We got that password: ', password) -- prints 'hunter2'
    end, 'IdrisSmells')
end, false)
```


server/functions.lua
```diff
@@ -196,8 +196,15 @@ function PaycheckInterval()
...
+-- Callback Functions --
 
+-- Client Callback
+function QBCore.Functions.TriggerClientCallback(name, source, cb, ...)
+    QBCore.ClientCallbacks[name] = cb
+    TriggerClientEvent('QBCore:Client:TriggerClientCallback', source, name, ...)
+end
+
+-- Server Callback
 function QBCore.Functions.CreateCallback(name, cb)
```
uShifty, line 208: I think you should change the previous callback natives which would make it a breaking commit to match the naming schema. TriggerServerCallback/CreateServerCallback. Other than that I cannot wait for this to get merged

ItsANoBrainer: I thought about that as well. Seems like it seemed like too large of a breaking commit to do so but will see what everyone things.

IdrisDose:
Looks fine.

However there was a discussion of changing the callback internals to have a unique id and name to prevent race conditions.

IdrisDose (approved):

xenfovn (approved):
NICE WORK

GhzGarage:
Has conflicts

PR #686, opened 2022-05-05
To remove decimals Money

merpindia:
**Describe Pull request**
It will remove decimal money from each and every transaction


xenfovn (approved):

Mojito-Fivem:
What if somebody wants decimal money?

marathivikya:
> What if somebody wants decimal money?

why u always funny?

PR #693, opened 2022-05-07
Lint support

TheiLLeniumStudios:
**Describe Pull request**


server/events.lua
```diff
@@ -209,40 +209,29 @@ QBCore.Functions.CreateCallback('QBCore:HasItem', function(source, cb, items, am
...
-                            retval = true
-                        end
-                    end
+                if item and item.amount >= v then
+                    count += 1
                 end
             else
-                finalcount = #items
                 local item = Player.Functions.GetItemByName(v)
                 if item then
```
IdrisDose, line 221: The logic doesn't make sense here.

IdrisDose (changes requested):
```lua
if item then
                    if amount then
                        if item.amount >= amount then
                            count += 1
                        end
                    else
                        count += 1
                    end
                end
```

the logic doesn't make sense on line 221 -> 229 of HasItem change

TheiLLeniumStudios:
> ```lua
> if item then
>                     if amount then
>                         if item.amount >= amount then
>                             count += 1
>                         end
>                     else
>                         count += 1
>                     end
>                 end
> ```
> 
> the logic doesn't make sense on line 221 -> 229 of HasItem change

Simplified the logic. It should make sense now @IdrisDose :)

server/functions.lua
```diff
@@ -63,7 +63,7 @@ end
 
 function QBCore.Functions.GetPlayers()
     local sources = {}
-    for k, v in pairs(QBCore.Players) do
+    for k, _ in pairs(QBCore.Players) do
```
BerkieBb, line 66: You can remove this underscore completely, it doesn't need to be there if it's not used.
In lua when you have a parameter that's not used after one that is used and there are none used after the one that's not used, then you can delete everything after the one that's used and has nothing that is used after it.

IdrisDose: It's a standard in Lua to use underscores as not-used variables.

BerkieBb: But instead of adding them, you can remove them if there is nothing else beong used after it, just like you can remove unused variables from the funtion arguments in events, no need for it

BerkieBb: I only use underscores when there is a variable after the underscore that's used, otherwise I remove it because it is something extra that serves no purpose basically

BerkieBb:
```suggestion
    for k in pairs(QBCore.Players) do
```

PR #702, opened 2022-05-15
FEAT: Allow getting offline Player object

mmoftah:
**Describe Pull request**
This PR allows developers to get an offline Player object and be able to update the player without needing to take on a database dependency. This enabled resources to not have to understand the rest of the player object or schema to update a small part of the player. A good example of this would be a boss menu that has to fire offline players. It can now simplify the logic and just check if the player is not online, then they could get the offline player object.

This PR also safeguards against some issues where source is used. This is mostly logging, and I would recommend a follow up PR by someone else to add logging. It was not important to me as other scripts should be logging what they are doing to the offline player, I think.

An example usage would look like:
```
  local Employee = QBCore.Functions.GetPlayerByCitizenId(target)
  local Online = true
  if not Employee then
    Employee = QBCore.Functions.GetOfflinePlayerByCitizenId(target)
    Online = false
  end

  ... logic that modifies Employee

  if not Online then
    Employee.Functions.Save()
  end
```

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [yes/no] (Be honest) Yes, to the best of my ability. 
- Does your code fit the style guidelines? [yes/no] To my knowledge, but I added extra spacing to make it super clear where the code is being added.
- Does your PR fit the contribution guidelines? [yes/no] Yes.


server/player.lua
```diff
@@ -36,12 +36,38 @@ function QBCore.Player.Login(source, citizenid, newData)
     end
 end
 
+function QBCore.Player.GetOfflineUser(citizenid)
```
BerkieBb, line 39: I think it's more fitting if you named it `GetOfflinePlayer`

BerkieBb (changes requested):

BerkieBb:
Also, your indentation is a bit off, in lua we use 4 spaces/tabs, you have 2 spaces in a lot of parts of the code

server/player.lua
mmoftah: Renamed! Good callout!

BerkieBb (approved):

server/player.lua
```diff
@@ -260,16 +300,20 @@ function QBCore.Player.CreatePlayer(PlayerData)
...
-        else
-            TriggerEvent('qb-log:server:CreateLog', 'playermoney', 'RemoveMoney', 'red', '**' .. GetPlayerName(self.PlayerData.source) .. ' (citizenid: ' .. self.PlayerData.citizenid .. ' | id: ' .. self.PlayerData.source .. ')** $' .. amount .. ' (' .. moneytype .. ') removed, new ' .. moneytype .. ' balance: ' .. self.PlayerData.money[moneytype] .. ' reason: ' .. reason)
-        end
-        TriggerClientEvent('hud:client:OnMoneyChange', self.PlayerData.source, moneytype, amount, true)
-        if moneytype == 'bank' then
-            TriggerClientEvent('qb-phone:client:RemoveBankMoney', self.PlayerData.source, amount)
+
+        if not self.Offline then
+            self.Functions.UpdatePlayerData()
+            if amount > 100000 then
```
nullvariable, line 306: is there a reason you wouldn't want logs for this if the action is offline? This is a pattern that continues throughout, I believe that for consistency we would want logs even if the user is offline. They just need to be tweaked so the `source` isn't included if the user is offline

PR #726, opened 2022-06-16
Fix ignoreList Compare in function GetPeds

ghost:
**Describe Pull request**

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [yes/no] (Be honest) yes
- Does your code fit the style guidelines? [yes/no] yes
- Does your PR fit the contribution guidelines? [yes/no] yes

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -183,8 +183,9 @@ function QBCore.Functions.GetPeds(ignoreList)
     for i = 1, #pedPool, 1 do
         local found = false
         for j = 1, #ignoreList, 1 do
-            if ignoreList[j] == pedPool[i] then
+            if ignoreList[j] == GetEntityModel(pedPool[i]) then
                 found = true
+                break
             end
         end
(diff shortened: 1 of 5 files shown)


GhzGarage:
You didn't fill out the template so can you tell me what issue this was causing that requires modifying it?

SuelTheDev:
> You didn't fill out the template so can you tell me what issue this was causing that requires modifying it?

Oh! Sorry. The function takes as parameters a hash list such as { `mp_f_freemode_01`, etc } of the peds must be ignored. But what is returned by GetGamePool is a list of "ids" of the entities so the comparison in the ignore list passed by the parameters would never give true for any item. So not ignoring what you want.

Lionh34rt:
If I remember correctly, ignorelist is supposed to be a list of peds and not ped models.

Use case: get closest ped that is not a player in qb-drugs:client:cornerselling

PR #757, opened 2022-07-18
Fix optin toggle and add SetPlayerData function to player object

BerkieBb:
**Describe Pull request**
This fixes the optin toggle by adding a function that can set the right data inside the playerdata, the SetPlayerData function is more useful than just for fixing this, it can make lots of operations easier and maybe more secure as well

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? [no] (Be honest)


tom-osborne (approved):
Tested ok.

PR #759, opened 2022-07-19
Update vehicles.lua 'sportsclassic' to 'sportsclassics'

nzkfc:
**Describe Pull request**
It seems there were sportsclassics and sportsclassic in the cateogies, after having a look around (and BerkieB suggested I check qb-customs) I see the category should be just 'sportsclassics'

Note: This may break peoples servers if they use configs that find 'sportsclassic', good example is okokvehicleshop where you set the categories. So might be worth mentioning that somewhere if important.


tom-osborne (changes requested):
Nice find, but you have made a spelling mistake with double "s" at the end. 

`'sportsclassicss'` should be `'sportsclassics'`

nzkfc:
> Nice find, but you have made a spelling mistake with double "s" at the end.
> 
> `'sportsclassicss'` should be `'sportsclassics'`

haha oh sheeet late night eyes, updated.

PR #762, opened 2022-07-21
Fix styling, docs, and a possible bug

D4isDAVID:
**Describe Pull request**
Looks like I missed some stuff in my previous PR, so here we go again.

In terms of styling, this:
- Ensures consistent indentation and spacing
- ~~Uses single quotes over backticks~~ (Invalid)

In terms of docs, this:
- Fixes an invalid JSDoc parameter
- Uses `nil` over `void` in Lua docs
- Adds a missing `nil` type for a parameter

And this also fixes a possible bug where the `Locale:replace` method would not clear and extend the locale correctly.

- Have you personally loaded this code into an updated qbcore project and checked all it's functionality? **no**
- Does your code fit the style guidelines? **yes**
- Does your PR fit the contribution guidelines? **yes**

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -244,7 +244,7 @@ function QBCore.Functions.IsWearingGloves()
     local ped = PlayerPedId()
     local armIndex = GetPedDrawableVariation(ped, 3)
     local model = GetEntityModel(ped)
-    if model == 'mp_m_freemode_01' then
+    if model == `mp_m_freemode_01` then
         if QBCore.Shared.MaleNoGloves[armIndex] then
             return false
         end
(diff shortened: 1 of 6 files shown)


client/functions.lua
```diff
@@ -244,7 +244,7 @@ function QBCore.Functions.IsWearingGloves()
     local ped = PlayerPedId()
     local armIndex = GetPedDrawableVariation(ped, 3)
     local model = GetEntityModel(ped)
-    if model == `mp_m_freemode_01` then
+    if model == 'mp_m_freemode_01' then
```
BerkieBb, line 247: Change this back, this will make an error because it will try to compare a number to a string, backticks are used to make hashkeys, which are numbers

BerkieBb (changes requested):

client/functions.lua
D4isDAVID: Sorry about that

BerkieBb (approved):

PR #764, opened 2022-07-23
Fixed isnetworked

Maykellll1:
**Describe Pull request**
With the above way, the condition was always true because (false or true) is true or (nil or true) is true. With this change it is fixed.


BerkieBb (approved):

jnccloud:
Does this fix the issue with vehicles becoming owned by the 'world', meaning the problem where you cannot delete or park them normally?

PR #781, opened 2022-08-07
get and set color types alongside vehicle properties 

LRemp:
**Describe Pull request**
This adds an option to set/get primary and secondary  vehicle color types, although **qb-customs** already has multiple color types available in it's paint list, but it utilizes hardcoded paint ids, whilst this will allow external resources to set color type to any color, including custom colors(invoked by **SetVehicleCustomPrimaryColour** or **SetVehicleCustomSecondaryColour**  natives).


client/functions.lua
```diff
@@ -443,6 +443,8 @@ function QBCore.Functions.GetVehicleProperties(vehicle)
     if DoesEntityExist(vehicle) then
         local pearlescentColor, wheelColor = GetVehicleExtraColours(vehicle)
 
+        local colorPrimaryType, _, _ = GetVehicleModColor_1(vehicle)
+        local colorSecondaryType, _ = GetVehicleModColor_2(vehicle)
```
BerkieBb, line 447: You can remove these underscores, there are no other variables after it being uses so it won't cause issues

BerkieBb (changes requested):

PR #792, opened 2022-08-18
Bypass for loading QBCore in FxDK

MannaMuncher:
FxDK does not include license identifiers by default and thus breaks scripts that utilize those. Without this you will get the "No valid Rockstar license" error when connecting. This fix was not my own I found it in the post linked below and tested myself.

https://forum.cfx.re/t/connection-rejected-by-server-fxdk/2504365/5


ChatDisabled:
Interesting problem indeed. Besides the function `QBCore.Functions.GetIdentifier` there are some instances where the native `GetPlayerIdentifiers` also gets used within QBCore. Definitely something the team should look into.

Update your branch though! Mind the linting issues as well @MannaMuncher.
alt text

D4isDAVID:
@ChatDisabled
The linting issues are most likely caused by the previous merge, so just updating the branch should fix it

MannaMuncher:
I was in the middle of moving but I fixed it now.

MannaMuncher:
You guys should be able to update branches from the PR view the merge blocking is unnecessary.

jeann2013 (approved):

D4isDAVID (approved):
LGTM

Z3rio (approved):
Code looks good to me, although I have not been able to test it myself with FXDK.
I will have to trust the other testers that have tried it with FXDK.

PR #812, opened 2022-08-25
feat: password prompt on server join (configurable) + maitenance mode

Spudgun:
**Describe Pull request**
Enables the use of a password prompt upon server join. This is configurable in the `config.lua` and also has a toggle/set event for toggling on/off and setting the password.

Further to this, I have created a Maintenance Mode, this is different from the Closed Server option currently. It allows for the server be to joined by people with permissions set in the `QBConfig.Server.ClosedWhitelist` table **AND** also anyone who knows the maintenance mode password.

I am planning on building this functionality into [qb-adminmenu](https://github.com/qbcore-framework/qb-adminmenu).

image


D4isDAVID (changes requested):
Seems to work fine, the only 2 complaints I have are:
1. It's probably better to ask for a password before checking for ban, allowance, etc.
2. After the password attempts you get kicked but then you can instantly connect and try again, maybe players should get a timeout of some time until trying again. (Maybe every kick the timeout gets longer until at some point you're banned?)

Z3rio:
Yeah, having the password in the config might not be the best since it can be dumped, although on second thought, all the people dumping the server / files would already have to know the password haha.
So the only issues with this PR would be that it is outdated & the stuff that david mentioned.

D4isDAVID:
> Yeah, having the password in the config might not be the best since it can be dumped, although on second thought, all the people dumping the server / files would already have to know the password haha. So the only issues with this PR would be that it is outdated & the stuff that david mentioned.

But that part is actually only loaded on the server so it isn't possible to dump, no?

Spudgun:
> > Yeah, having the password in the config might not be the best since it can be dumped, although on second thought, all the people dumping the server / files would already have to know the password haha. So the only issues with this PR would be that it is outdated & the stuff that david mentioned.
> 
> But that part is actually only loaded on the server so it isn't possible to dump, no?

This was my understanding as well. However, I have been away from the FiveM Developer scene for a while so unsure if this has changed.

> Seems to work fine, the only 2 complaints I have are:
> 
> It's probably better to ask for a password before checking for ban, allowance, etc.
After the password attempts you get kicked but then you can instantly connect and try again, maybe players should get a timeout of some time until trying again. (Maybe every kick the timeout gets longer until at some point you're banned?)

Regarding this, it is definitely possible to add this functionality and also a feature regarding the timeout/banning.

Spudgun:
> Seems to work fine, the only 2 complaints I have are:
> 
> 1. It's probably better to ask for a password before checking for ban, allowance, etc.
> 2. After the password attempts you get kicked but then you can instantly connect and try again, maybe players should get a timeout of some time until trying again. (Maybe every kick the timeout gets longer until at some point you're banned?)

These changes have all been implemented, excluding the ban check before a password input.

r0adra93 (approved):

PR #814, opened 2022-08-26
feat: some useful functions + command permission change

Spudgun:
**Command Permissions**
Simplify command permissions to be one parameter, as opposed to an infinite amount. I believe this is a lot cleaner and simpler way of doing permissions. 

I've also searched the entire framework for `QBCore.Commands.Add` function and no script is utilizing more than one parameter in this function. So it won't have any conflicts with the base framework or issues.

**Useful Functions** 
Adds the ability to manipulate tables (copy, copy strict, combine, and search) and grouping of digits together.

`QBShared.GroupDigits(1000000, true)` -> **$1,000,000**
`QBShared.GroupDigits(1000000, false)` -> **1,000,000**


Waves966 (commented):
If you're using lua 5.4 you could also just do table.copy(), I don't see much use for a function like this to be added to the core.

Spudgun:
> If you're using lua 5.4 you could also just do table.copy(), I don't see much use for a function like this to be added to the core.

Was unaware this functionality was included with lua 5.4. Can easily remove that function if this is the case.

Thoughts on the rest of the features???

Spudgun:
> If you're using lua 5.4 you could also just do table.copy(), I don't see much use for a function like this to be added to the core.

So lua 5.4 is enabled in the `fxmanifest.lua` for qb-core but when using `table.copy(tbl)` it doesn't seem to work and throws a CLI error of

`SCRIPT ERROR: @qb-core/server/commands.lua:65: attempt to call a nil value (field 'copy')`

Am I missing something here or is copy not a function of the `table` object with Lua 5.4?

Spudgun:
refactored the permission change to ensure backward compatibility for extra perms, if supplied as perm @tom-osborne request.

r0adra93:
update the branch and see if it conflicts with the base branch.

PR #822, opened 2022-09-01
Update de.lua

CptNoHand:
sounds better like this


Z3rio (approved):
Looks good to me, if someone that actually knows this language can confirm that it is better then that would be good.
I dont see any issues with the code or such though, just unsure if its better then the original or not :)

TheRealCasiv2 (changes requested):
Looks good!
But i suggest changing the duty messages to:
"Du befindest dich nun im Dienst!" and "Du befindest dich nun nicht mehr im Dienst!"

But otherwise it looks really good

TheRealCasiv2 (changes requested):
It's better but still has a grammatical error, correct would be like i said: "Du befindest dich nun im Dienst!" and "Du befindest dich nun nicht mehr im Dienst!". So replace the "jetzt" with "nun"

Then it would be correct.

TheRealCasiv2 (approved):
It's correct now. 
I Approve.

PR #825, opened 2022-09-02
Removed Weaponized Vehicle

Ghosty786:
Removed Oppressor
Removed Stromberg
Removed Deluxo
Removed Ruiner 2000

## Checklist


GhzGarage:
There’s no reason to delete the code when it can be commented instead

Ghosty786:
> There’s no reason to delete the code when it can be commented instead
Just Removed in-case for someone new who doesn't know, but you are right we could comment it out.

PR #828, opened 2022-09-05
Framework Optimization

KamuiKody:
for optimization of other resources.

## Description

This is to further optimize other resources that commonly use loops to check if in vehicle. This also gathers common variables so all these resources wont have to call for the same variables. 

RegisterNetEvent('QBCore:Client:EnteringVehicle')
triggers once when you try to enter a vehicle
to break a loop triggered by this you would add this line to it:
if GetVehiclePedIsTryingToEnter(PlayerPedId()) == 0 then break end

RegisterNetEvent('QBCore:Client:AbortVehicleEntering')
triggers once when you abort entering a vehicle or the door is locked.
it passes no variables since typically it would be used to kill a loop

RegisterNetEvent('QBCore:Client:EnteredVehicle')
triggers once when you enter a vehicle or when you spawn inside of a vehicle
if IsPedInAnyVehicle(PlayerPedId()) == 0 then break end

RegisterNetEvent('QBCore:Client:LeaveVehicle')
triggers once when you leave a vehicle

## Checklist


sinphoenix (approved):
Ran it my server, works fine!

r0adra93:
works

r0adra93 (approved):

Waves966 (commented):
Mainly the problem with using this is the fact that baseevents is an outdated standard which is apart of the outdated cfx-server-data repository and most likely won't be included in the updated integrated solutions. Probably best not to rely too much on the base cfx resources at this point as the cfx-server-data is considered deprecated at this point.

KamuiKody:
Not to be combatative but id like to know where you saw this because i actually got these events front the cfx docs and upon search couldnt find anything on the forums about them except for the fact that they are used as a tool for developement.

Waves966:
https://github.com/citizenfx/cfx-server-data

KamuiKody:
I see. They said in the reafme that it would be moved they didnt say when though and if they were going to fully remove it. Which they should i guess the way i saw it was these events come on a loop in a base server anyway theres no reason for us to call a loop in 75% of our resources to look for a vehicle. If you do have a better suggestion im all ears for it ill test it out a bit

mahanmoulaei:
> If you do have a better suggestion im all ears for it ill test it out a bit

In order to avoid relying on baseevents which runs every frame, you could create a loop with a wait of 200ms and trigger an event on vehicle change or even seat change.

KamuiKody:
but then you are asking people who may or may not be developers to remove something for server performance where as this update utilizes a function that already exists. im going to be looking into low level game events. also  making something call every 200ms vs every frame does not give u substancial performance gain. The low level game events however should remove the loop completely. So whenever they do depricate baseevents completely there should be better options.

Irishstevie (approved):
I have been running this for a month and installed it on a few new copy of the core
Its all good to go

PR #837, opened 2022-09-13
fix(readme): Acknowledge Proper Licence Ownership

Mycroft-Studios:
## Description

Acknowledge the fact that the code is modified from ESX,
Incase any of you shalt to think otherwise, heres a picture of your lord Kakarot admitting that it is a modified es extended, therefore your code must legally follow our licencing, thanks
https://cdn.discordapp.com/attachments/844240881283366962/1019029136506888294/IMG_1766.png

This was removed For Some Reason in:
https://github.com/qbcore-framework/qb-core/commit/b614e703b26261d666b93307f762d7407f0baf34

Thanks For Reading,
Mycroft


HunterHoch (approved):
**_This is the correct thing to do._**

GhzGarage:
This was already resolved with Ark and Technetium long ago

ArkSeyonet:
I haven't looked at the code, but I'm assuming that if the license information was removed, then that means you successfully removed the portions of ESX code to where it no longer contains code that belongs to the ESX copyright.

GhzGarage:
I only ever removed it because i got permission from you once it was changed. Not sure what you mean on the monetization thing though, i've always stayed true to who i am from the beginning and always done things by the book

ArkSeyonet:
It is just a word of caution, from me to you. If you are keeping on the straight and narrow, good. If not, straighten up because I vouched for you my dude. I haven't kept up with you lately, so we don't really know each other that well anymore, so just giving a precautionary word between you and myself.

I cannot say that I remember giving you permission to remove the license information completely. I thought we talked about you working on rewriting the ESX parts, so that you could remove the ESX license information from those parts of the code. I never received word from you that you had completed that process or anything, as far as I remember, since I took a break from FiveM while that process was going on.

PR #858, opened 2022-09-27
chore: recommend `cfxlua-vscode` extension rather than `fivem-vscode`

D4isDAVID:
## Description

[overextended/cfxlua-vscode](https://github.com/overextended/cfxlua-vscode) was recently released as [a VSCode extension](https://marketplace.visualstudio.com/items?itemName=overextended.cfxlua-vscode) which works on top of [lua-language-server](https://github.com/sumneko/lua-language-server), making it better to develop with rather than [fivem-vscode](https://marketplace.visualstudio.com/items?itemName=fivem-vscode.fivem-vscode).


Z3rio (approved):
Looks good to me. This is indeed a great addon to qb-core

PR #826, opened 2022-09-03
Add locale convar and locale fallback

D4isDAVID:
## Description

Add support for using the `qb_locale` convar added by qbcore-framework/txAdminRecipe@04a7be2, since #520 seems abandoned.
Add a new `fallbackLang` option for locales, where if a translation is not found it is taken from the fallback.
Fix a bug where `warnOnMissing` couldn't be set to `false`.

Closes #520
Closes #581

## Checklist

- [x] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [x] My code fits the style guidelines.
- [x] My PR fits the contribution guidelines.

PR #860, opened 2022-10-01
fix: locale instances not actually being created

D4isDAVID:
## Description

Fix Locale instances not actually being created.

After #693, calling `Locale:new` would not actually create a new table for the Locale instance but instead use the already existing `Locale` variable itself, which would result in running the following code:
```lua
setmetatable(Locale, Locale)
```

This would cause an error when trying to index the table with keys that don't exist in `Locale`.
unknown

With the intended way that `Locale` is used, this PR should work fine and be backwards compatible.

## Checklist

### DIFF
--- a/shared/locale.lua
+++ b/shared/locale.lua
@@ -30,10 +30,10 @@ end
 --- Constructor function for a new Locale class instance
 --- @param opts table<string, any> - Constructor opts param
 --- @return Locale
-function Locale:new(opts)
-    setmetatable(self, Locale)
+function Locale.new(_, opts)
+    local self = setmetatable({}, Locale)
 
-    self.fallback = opts.fallbackLang and Locale.new({}, {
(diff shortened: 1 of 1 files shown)


Z3rio:
I believe this issue is outdated since no one gets this error anymore.

D4isDAVID:
@Z3rio
As an example take `Locale.whatever`.

As seen in [shared/locale.lua](https://github.com/qbcore-framework/qb-core/blob/312f24d245a6cf66d45f9f4fcb2f52838a2b361c/shared/locale.lua), this is never given a value, but still the expected behavior is that it simply has a value of `nil`.

But when you actually try using it, an error will be thrown, which is (at least in my opinion) most definitely unexpected behavior. It was quite annoying to debug when trying to implement #826.

Z3rio (approved):
Everything seems to be working fine after testing this.
Great PR.

PR #877, opened 2022-10-25
🐛  Adding Logs For /me as it doesnt come default with the event for it in commands.lua 🐛 

Marshxan:
## Adding 

so by default.... the ``qb-smallresources/server/logs.lua`` there is a ``['me'] = '',`` logs.... by default
and in the file ``qb-core/server/commands.lua`` at the bottom there is no logs export and this may annoy some people that wants logs... and they will open a support thread in support and this will happen over and over again and people will have to keep asking for the event to paste it in!

Discord: Preview: https://part-of-the.cutecatgirl.club/BtEj0KtY7abb
Alive: Preview: https://part-of-the.cutecatgirl.club/HXoanshrOmXF
Dead: Preview: https://part-of-the.cutecatgirl.club/EvY9ARlgBQZg

### DIFF
--- a/server/commands.lua
+++ b/server/commands.lua
@@ -290,7 +290,7 @@ QBCore.Commands.Add('me', Lang:t("command.me.help"), {{name = Lang:t("command.me
         local tCoords = GetEntityCoords(target)
         if target == ped or #(pCoords - tCoords) < 20 then
             TriggerClientEvent('QBCore:Command:ShowMe3D', Player, source, msg)
-            TriggerEvent('qb-log:server:CreateLog', 'me', 'Me Command Log', 'white', '**ID: ' .. source .. ') Message:** ' .. msg, false)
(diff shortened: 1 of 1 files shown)


server/commands.lua
```diff
@@ -288,8 +288,11 @@ QBCore.Commands.Add('me', Lang:t("command.me.help"), {{name = Lang:t("command.me
         local Player = Players[i]
         local target = GetPlayerPed(Player)
         local tCoords = GetEntityCoords(target)
+        local Player = QBCore.Functions.GetPlayer(source)
+        local citizenID = Player.PlayerData.citizenid
```
tom-osborne, line 292: Please move both `Player` and `citizenID` outside of the for loop, since we only need to get this data once.

tom-osborne (changes requested):

PR #887, opened 2022-12-07
[ENHANCEMENT] Added updateItemMetaData function

TGIANN:
## Description

With this function, you can update the item metada without deleting the items

## Checklist


server/player.lua
```diff
@@ -447,6 +447,17 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
         return false
     end
 
+    function self.Functions.updateItemMetaData(item, slot, info)
```
Z3rio, line 450:
```suggestion
    function self.Functions.UpdateItemMetaData(item, slot, info)
```
Please follow the "format standards" in qb-core. Considering all other functions start with capital letters, this should also start with an capital letter :))

server/player.lua
```diff
@@ -447,6 +447,17 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
...
+		if not slot then return end
+		if not info then return end
```
Z3rio, line 452:
```suggestion
		if not slot or not info then 
		    return 
		end
```

Better readability

Z3rio (changes requested):

Z3rio:
From what I know, all inventory functions have been moved to qb-inventory instead of qb-core.
I believe this should also be in there if you want to add it. Otherwise I see no real issue with the code.
I have yet to test it ingame though :)

PR #899, opened 2022-12-21
🎨 Refactor: Refactors jobs conditional statement

Qwerty1Verified:
## Description

Simply a refactor to the recent change for qb-jobs, instead of indenting the whole table with a "not" condition, just check for the condition above and return if the condition is true.

## Checklist


Z3rio (approved):
LGTM

PR #893, opened 2022-12-14
[FIX] shared/vehicles.lua car names & other issues

SirOMGitsYOU:
## Description

This fixes a few grammatical and errors amongst the vehicles.lua to bring it back in-line and correct with the current vehicle set

## Checklist

- [x] Fixed error on ASBO shop tag
- [x] Fixed Duplicate vehicle names
- [x] Added missing Manchez
- [x] Fixed incorrect vehicle name.
- [x] Fixed boats being called under their brand name rather than the actual name of the boat
- [x] Fixed name for Ratloader

PR #901, opened 2022-12-22
[FIX] more shared/vehicles.lua car names & brands

SirOMGitsYOU:
## Description

This fixes a few grammatical and errors amongst the vehicles.lua to bring it back in-line and correct with the current vehicle set and fixed a few branding name issues

This is the final PR for this one, All vehicle names and brands are correct from what I can tell :) Sorry the final commits couldn't be included in https://github.com/qbcore-framework/qb-core/pull/893, I got covid so it kinda delayed my updates but back now and was able to go through every single vehicle.


SirOMGitsYOU:
That's everything all ready for review 👍🏻

Z3rio (approved):
Great work like last time, looks good to me.

PR #905, opened 2022-12-27
Small Fixes: QBCore.Debug and Spelling Mistake

jellyton69:


Z3rio (changes requested):
Put the color in the config of qb-core instead, thank you.

PR #908, opened 2022-12-28
Departments

Bilalgaming250:
Department System to Jobs 

Adds the support for jobs to have "departments" example for this would be that Police job can have different departments
and you can have different colors for trackers or etc. Basically So you can differentiate departments in jobs.


ETHX0018:
I think this feature will help the developers make something interesting

PR #911, opened 2022-12-31
Added back QBCore:Server:SetMetaData

erikmeyer08:
Added back the event QBCore:Server:SetMetaData


i-kulgu (commented):
This function is deprecated, use Player.Functions.SetMetaData instead.

PR #913, opened 2023-01-04
Add CreateBlip function

KamuiKody:
allows more efficient blip creation

## Description

My pull request adds a function to the core to create a blip and store it within the core object. This will come in handy as many different resources use this exact same code. This also will prevent the server owner from going over the legend count that bugs out then map.

local blip = QBCore.Functions.CreateBlip('unique_id', vector3(0,0,0), 1--[[sprite]], 0.6--[[scale]], 4--[[display]], 1--[[color]], true--[[shortRange]], 'Legend Label' --[[label]])

## Checklist

- [ x] My code fits the style guidelines.


r0adra93 (approved):
Works Great SEND IT!

jellyton69 (changes requested):
Use the following instead

```lua
function QBCore.Functions.CreateBlip(name, coord, sprite, scale, display, color, shortRange, label)
    QBCore.Blips[name] = AddBlipForCoord(coord.x, coord.y, coord.z)
    SetBlipSprite(QBCore.Blips[name], sprite)
    SetBlipScale(QBCore.Blips[name], scale)
    SetBlipDisplay(QBCore.Blips[name], display)
    SetBlipColour(QBCore.Blips[name], color)
    SetBlipAsShortRange(QBCore.Blips[name], shortRange)
    if #QBCore.Blips >= 100 and label then print('ERROR: Too Many Blips In Legend') return QBCore.Blips[name] end
    if not label then return QBCore.Blips[name] end
    AddTextEntry(label, label)
    BeginTextCommandSetBlipName(label)
    EndTextCommandSetBlipName(QBCore.Blips[name])
    return QBCore.Blips[name]
end
```

Using AddTextEntry bypasses the string limit, so you don't get flashing blips in the map.

jellyton69:
Also I'm not 100% sure but I think the blip limit is 99, not 100

Enz0Z:
Hey! I think `#QBCore.Blips >= 100` may not work, because you are using a key, value table type, and not an array.

i-kulgu (approved):
Tested it out and it works like supposed.

TreP-coder:
yea you're only allowed 99 blips. i would try to keep it in 70 range

PR #938, opened 2023-01-31
Fix drawtext error in html

worapolburaphan:
1. Start QB-Core on a slow or low-spec computer
2. Access the Debug UI page (localhost:13172)
3. Encounter an error in the drawtext.js file
4. The issue stems from the script loading successfully but any resource call function (hideText) and sending an event message before index.html of all resources are fully loaded on NUI it make
```html
 <div className="text" />
```
Doesn't loaded.

## Checklist


zwkf (approved):
LGTM

PR #948, opened 2023-02-18
New transaction method and new online field in database

PatalJunior:
## Description

Add Transaction of money using Mysql Transactions and, new online field in database (mainly useful for web admin panels)
Using xpcalls in Callbacks to aid debugging, since without it, it's harder to know where the error origin is.

## Checklist

- [✔️] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [✔️] My code fits the style guidelines.
- [✔️] My PR fits the contribution guidelines.


server/player.lua
```diff
@@ -346,12 +410,11 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
         local difference = amount - self.PlayerData.money[moneytype]
         self.PlayerData.money[moneytype] = amount
 
-        if not self.Offline then
```
zwkf, line 349: Is this changed intentionally?

server/player.lua
```diff
@@ -402,15 +465,15 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
         self[fieldName] = data
     end
 
-    if self.Offline then
```
zwkf, line 405: Why was this changed?

zwkf (changes requested):
Lots of unrelated changes which are outside of the scope of this pr

server/player.lua
PatalJunior: That was changed due the Offline logic changing so I could add it onto the database.
I've tested it all and works, not unrelated to pr, since it's needed to add the online field.

PatalJunior: For more context, the way it was synced and stored before didn't allow it to properly sync onto the database, and most of the times we want to know if player is online and not offline, so I figured since I was reworking the code I changed it to online, inside PlayerData, that way we can know specifically what character is online.
Old way would only let us know if the player was online but not the character itself.

PR #957, opened 2023-03-01
enhancement: Add documentation (luadoc) to functions

tom-osborne:
## Description

This PR adds documentation to server functions (luadoc).

## Checklist


server/functions.lua
```diff
@@ -8,12 +8,19 @@ QBCore.UsableItems = {}
 -- ex: local player = QBCore.Functions.GetPlayer(source)
 -- ex: local example = player.Functions.functionname(parameter)
 
+---Gets the coordinates of an entity
```
zwkf, line 11: Do these comments need to have `---` or `--`?

See: https://github.com/lunarmodules/LDoc

server/functions.lua
```diff
@@ -234,39 +286,62 @@ end
...
 function QBCore.Functions.TriggerCallback(name, source, cb, ...)
     if not QBCore.ServerCallbacks[name] then return end
     QBCore.ServerCallbacks[name](source, cb, ...)
 end
 
 -- Items
 
+---Create a usable item
+---@param item string
+---@param data function
```
zwkf, line 320: Is `function` correct here? Or would this be a table?

tom-osborne: The usage for this function is to add a callback function as the second parameter. I did consider changing the argument to `cb` like it used to be.

It was changed from `cb` to `data` here: https://github.com/qbcore-framework/qb-core/pull/813

PR #962, opened 2023-03-11
🐛 fix in character remove

Veolene:
This addition in code is to remove the apartment along with the character. This is to avoid a new character that take over the removed citizenid's apartment. This will, for example, avoid multicharacter skipping the character creation process.

## Description
Currently, when you are removing a character, the apartment attached to the CID will remain in the database. Whenever someone creates a new character, it will over the removed CID they also take over that CID's apartment. 

This is a problem in multicharacter for example, because the script will think this new character already owns an apartment and will skip a majority of the character creation process. Removing the apartment that is attached to the CID in apartments along with the character attached to the CID in players will make sure other scripts will work as intended.

## Checklist
- [X] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [X] My code fits the style guidelines.
- [X] My PR fits the contribution guidelines.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -539,13 +539,16 @@ function QBCore.Player.DeleteCharacter(source, citizenid)
     local result = MySQL.scalar.await('SELECT license FROM players where citizenid = ?', { citizenid })
     if license == result then
         local query = "DELETE FROM %s WHERE citizenid = ?"
-        local tableCount = #playertables
+        local tableCount = #playertables + 1 -- Add 1 for the apartments table
         local queries = table.create(tableCount, 0)
 
-        for i = 1, tableCount do
+        for i = 1, #playertables do
             local v = playertables[i]
             queries[i] = {query = query:format(v.table), values = { citizenid }}
         end
+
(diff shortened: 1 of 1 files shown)


server/player.lua
```diff
@@ -539,14 +539,17 @@ function QBCore.Player.DeleteCharacter(source, citizenid)
     local result = MySQL.scalar.await('SELECT license FROM players where citizenid = ?', { citizenid })
     if license == result then
         local query = "DELETE FROM %s WHERE citizenid = ?"
-        local tableCount = #playertables
+        local tableCount = #playertables + 1 -- Add 1 for the apartments table
```
dhawton, line 542: There's no reason for this. Your PR only requires 1 line addition.. add `apartments` to the playertables (https://github.com/qbcore-framework/qb-core/pull/962/files#diff-7bac2e2b00ff0a0b87f47152e9794e29e49043b171068b6d93fd65572dabac77R522) as the rest of it requires no code changes (the `citizenid` column is the same as does the query format).

PR #963, opened 2023-03-12
QBCore.Functions.DrawText Update

Inspector-Parrot:
The changes made to QBCore.Functions.DrawText were done to optimize the function and make it more readable and maintainable.

## Description

- Removed unnecessary function calls: The original function called several functions that were not necessary, such as `SetTextProportional(0)` and `SetTextDropShadow()`. These functions do not affect the behavior of the function and can be safely removed, reducing the amount of code that needs to be executed.

- Added default values: The original function did not have default values for any of the parameters, which meant that all parameters had to be specified every time the function was called. I added default values for the width and height parameters, which are set to 0.1 and 0.05, respectively. This means that if these parameters are not specified, the function will use these default values instead.

- Improved code readability: I made several minor changes to the code to improve its readability. The original function had the parameters in the following order: `x, y, width, height, scale, r, g, b, a, text`. While this order works, it can be confusing to remember the order of the parameters, especially if you are not using the function frequently. To make the function more intuitive, I changed the order of the parameters to: `text, x, y, scale, r, g, b, a, width, height`. This makes it easier to remember which parameter

## Checklist

- [X] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [X] My code fits the style guidelines.

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -18,7 +18,7 @@ end
 
 -- Utility
 
-function QBCore.Functions.DrawText(x, y, width, height, scale, r, g, b, a, text)
+function QBCore.Functions.DrawText(text, x, y, scale, r, g, b, a, width, height)
     SetTextFont(4)
     SetTextScale(scale, scale)
     SetTextColour(r, g, b, a)
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -19,15 +19,9 @@ end
 -- Utility
 
 function QBCore.Functions.DrawText(x, y, width, height, scale, r, g, b, a, text)
-    -- Use local function instead
     SetTextFont(4)
-    SetTextProportional(0)
     SetTextScale(scale, scale)
     SetTextColour(r, g, b, a)
-    SetTextDropShadow(0, 0, 0, 0, 255)
-    SetTextEdge(2, 0, 0, 0, 255)
-    SetTextDropShadow()
-    SetTextOutline()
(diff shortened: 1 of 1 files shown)


tom-osborne:
Nice PR, lovely description.

However, while I agree that changing the order of parameters does give an improvement to the function readability, this is a breaking change. Therefore, some level of impact analysis should be conducted prior to merging. Did you consider this at all?

Inspector-Parrot:
> However, while I agree that changing the order of parameters does give an improvement to the function readability, this is a breaking change. Therefore, some level of impact analysis should be conducted prior to merging. Did you consider this at all?

I'm not sure I'm understanding exactly what you mean. But I've been testing it on an updated qb base and everything seems to work fine. Anyways, the order of the parameters is not the reason for the pull request. The main reason is by removing unnecessary function calls to optimize the function and adding factory values ​​for some parameters so they don't have to be specified every time the function is called. 🦜

tom-osborne:
What I mean is any existing code that utilises this function will need to be updated to use the new order of parameters. This may be an insignificant amount of scripts in the base QBCore.

It may also have an effect on a large number of third-party resources. Third-party script creators or end users will need to be notified of the change too as it could start throwing lots of errors into people's consoles.

Inspector-Parrot:
Yeah i know what you saying now. Then is not needed to merge the part where the order of the parameters changed. Just merge the part where i removed unnecessary function calls

PR #977, opened 2023-04-28
% Symbol appear in garage menu

Andyauk:
fixes removed  %

## Description

fixes issues reported in bug report #976
after removed from latest update.

## Checklist

- [x] I have personally loaded this code into an updated qbcore project and checked all of its functionality. yes
- [x] My code fits the style guidelines. yes
- [x] My PR fits the contribution guidelines. yes

PR #979, opened 2023-04-29
Fixed % Sign In Notify

djontop:
Fixed % Sign In Notify We Get % On /job Etc. It Has Been Fixed. % Sign Was Missing From locale.lua.

### DIFF
--- a/shared/locale.lua
+++ b/shared/locale.lua
@@ -20,7 +20,7 @@ local function translateKey(phrase, subs)
 
     -- Initial Scan over result looking for substituions
     for k, v in pairs(subs) do
-        local templateToFind = '%{' .. k .. '}'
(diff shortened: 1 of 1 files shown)


Andyauk:
> Fixed % Sign In Notify We Get % On /job Etc. It Has Been Fixed. % Sign Was Missing From locale.lua.

already a PR in #977

djontop:
> > Fixed % Sign In Notify We Get % On /job Etc. It Has Been Fixed. % Sign Was Missing From locale.lua.
> 
> already a PR in #977

He not mentioned the issue. He just label like update locale lua nothing else I think it's different. 😄 You can remove this one

lazarospsa (approved):

PR #988, opened 2023-06-07
fix shared update

hakos47:
It's a temporary fix while we see how to do it better. but this in principle fixes the error that when a client entered it did not get the table of items etc updated. since shared_scripts does not serve to update information on both sides (client, server) it would fix problems with addItems, addItem, etc.

It is not the best solution. but it's a temporary fix

## Checklist

- [ x] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [ x] My code fits the style guidelines.


amantu-qbit:
Great work

amantu-qbit:
Also, why don't just push the whole QBshared table to client?

hakos47:
> Also, why don't just push the whole QBshared table to client?

The idea is good, but I think we should review it in more detail, as it could be an excessive burden. We should consider more efficient alternatives.

hakos47:
> Also, why don't just push the whole QBshared table to client?

I've changed the system and made it receive the information upon connection. I believe it works better than the previous one, at least in the tests I've conducted, and it fixes all the problems with job updates, vehicles, etc.

sirhvd:
I think you forgot to add "**src**" when using TriggerClientEvent.
`TriggerClientEvent('QBCore:Client:SharedUpdate', src, QBCore.Shared)`

server/events.lua
```diff
@@ -99,6 +99,7 @@ local function onPlayerConnecting(name, _, deferrals)
     end
 
     -- Add any additional defferals you may need!
+    TriggerClientEvent('QBCore:Client:SharedUpdate', QBCore.Shared)
```
GhzGarage, line 102: Missing source

PR #998, opened 2023-06-29
Attempt to improve set and get vehicle properties

luxu-gg:
## Why ?

Set and Get vehicle properties functions have missing mods and states, making it so that certain modifications will not be saved in the database.
While developing a tuning script I had to modify these functions in order to be able to save any kind of modification done to the vehicle.


GhzGarage:
Could you lessen the amount of changes made please

PR #1009, opened 2023-08-09
Remove unnecessary event

pushkart2:
The events triggered on duty change are for job update. But in reality job is never changed. This results in redundant events triggered and unnecessary data being used by other scripts like job scripts which can lag the client and server.

## Checklist


Qwerty1Verified:
What makes the events unnecessary?

PR #1012, opened 2023-08-18
Update CreatePlayer Function

Cocodrulo:
## Description

This pull request is for adding a more comprehensive way of notifying a player and checking if a player has items from the server side. For developers, specifically the ones that are starting, it would be easier to think that to notify a player or check if it has a specific item, you will first have to instance the Player variable.

## Checklist


Edvo1901:
Hmm, in my opinion, I doubt this is necessary. I am not saying it useless but let take the `HasItem` function for example. By default, I need only 1 line of code to check if a player has an item.

```lua
local hasItem = QBCore.Functions.HasItem(source, "item", "amount")
```

In your PR, I need to do up to 3 lines.

```lua
local Player = QBCore.Functions.GetPlayer(source)
if not Player then return end --This is optional
local hasItem = Player.Functions.HasItem("item", "amount")
```

Is this really necessary tho?

Cocodrulo:
I see your point, but most times yo check if a player has an item server side, you are the going to remove the item or do smth so the Player is already initialized. This change is not meant to be neccessary, because it is not, but is meant to help people that are starting with QBCore understand the things better, like you have the get the player you want to notify seems very logical to me, but in qbcore is not like that, and also it would be a change that doesn't break any script, but it is helpful for developers, that won't have to get the player source each time the want to check items o notify

Roester1 (approved):

PR #1013, opened 2023-08-24
Fix invalid increment syntax in `GetDutyCount()` and `GetPlayersOnDuty()`

BryceRussell:
## Description

Fix invalid syntax when incrementing `count` inside `GetPlayersOnDuty()` and `GetDutyCount()`

Had a few scripts throwing errors when calling this function, switching to `x = x + 1` syntax fixed all the console errors I was receiving

## Checklist

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -115,7 +115,7 @@ function QBCore.Functions.GetPlayersOnDuty(job)
         if Player.PlayerData.job.name == job then
             if Player.PlayerData.job.onduty then
                 players[#players + 1] = src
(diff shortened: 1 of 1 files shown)


Sharkiiie:
The original syntax is valid.
FiveM uses a custom Lua runtime which allows this operator.

BryceRussell:
I had a some server errors that referenced this file and this fixed seemed to stop the errors. I tried reproducing this again on a fresh server but I couldn't get it to error so I guess it something to do with my server, my bad!

PR #1051, opened 2023-12-01
Backwards compatibility for vehicles.lua

xDope7137:
## Description

you can just drag and drop the vehicles from older vehicles.lua now

## Checklist


MoskalykA:
bad idea https://stackoverflow.com/questions/71276468/what-is-the-fastest-way-to-go-through-an-array-table-with-numeric-indices

xDope7137:
> bad idea https://stackoverflow.com/questions/71276468/what-is-the-fastest-way-to-go-through-an-array-table-with-numeric-indices

buddy read the whole topic. they are enumerating 100,000+ values. People do not have more than 500 vehicles in that file.

by that logic, nothing should be using ipairs. It does not sense to make it work for all small sized tables.

Besides, are you going to let all server owners to convert their whole car tables to new layout? To save 0.2ms compute time? Just one time every restart? smh

xDope7137:
image

MoskalykA:
> you can just drag and drop the vehicles from older vehicles.lua now

the micro-optimisation is already done, so your argument doesn't make sense.

xDope7137:
qb dead?

xDope7137:
qb dead dead

PR #1073, opened 2024-01-17
FIX: Special characters in player name causes SQL error

1B0Y:
## Description
Pull request updates PlayerData.name under server/player.lua line 87, and simply parses everything but alphanumeric.
This simply fixes a bug that causes a SQL error being thrown when players use steam name with special characters, such as "ꜱʜᴀᴅʏ𝘅"

## Checklist
- [x ] My code fits the style guidelines.


TheNeo2k:
wouldnt;
`gsub("%S","")`

work better? Correct me if I'm wrong, your change gets rid of whitespace?

1B0Y:
Not in this case, as names such as ꜱʜᴅʏ𝘅_𝘅 (odd font used) would break SQL.
The gsub simply removes these

PR #1079, opened 2024-01-30
Feat: Add statebag for instances

FjamZoo:
## Description
Currently there's no way to tell whenever your client switched routing bucket, and a getter for which routing bucket you are in. Therefor this PR aims to add a feature which is already present in other frameworks this will help developers to easier support for qb-core throughout scripts, which requires this.

Usage
```lua
local playerId = PlayerId()

AddStateBagChangeHandler('instance', ('player:%s'):format(playerId), function(_, _, bucket)
    print("You just entered bucket", bucket)
end)
```

## Checklist


ChristianBDev (approved):
Looks good to me - Tested

PR #1095, opened 2024-03-24
Handle vehicle spawning and deleting commands on the server

EinS4ckZwiebeln:
## Description

The `/car [vehicleModel]` & `/dv` commands will now spawn and delete entities using exclusively server-side logic. This improves security and compatibility with the `sv_entityLockdown strict` mode.

## Checklist


GhzGarage:
The car command is cool but the dv i'd still like the functionality of deleting the closest vehicles too as it does now

EinS4ckZwiebeln:
@GhzGarage
A bit late but that should do the trick. :)

server/commands.lua
```diff
@@ -184,11 +184,23 @@ end, 'admin')
 -- Vehicle
 
 QBCore.Commands.Add('car', Lang:t('command.car.help'), { { name = Lang:t('command.car.params.model.name'), help = Lang:t('command.car.params.model.help') } }, true, function(source, args)
-    TriggerClientEvent('QBCore:Command:SpawnVehicle', source, args[1])
+    QBCore.Functions.SpawnVehicle(source, args[1], GetEntityCoords(GetPlayerPed(source)), true)
```
Mycroft-Studios, line 187: Doing it this way, will still spawn the vehicle on the client, regating any reason to have the PR.
`CreateVehicle` is an RPC native, which means it just internally sends a client event to spawn it, rather than the script making an event for it.

For correct server side spawning, `QBCore.Functions.CreateVehicle` should be used

Mycroft-Studios (changes requested):

server/commands.lua
EinS4ckZwiebeln: I see what you mean, and I agree, but it seems weird to me that this RPC-Native approach works with `sv_entityLockdown strict` just fine. I assume the official docs for this are heavily outdated, since the RPC-Client shouldn't be able to create the entity at all in strict mode, but it still does.

EinS4ckZwiebeln: Another thing to consider when using the `CreateVehicleServerSetter ` function is that the type of the vehicle is required, which cannot (at least not to my knowledge) be inferred from the model hash on the server side. You would need some dirty workarounds to make this work.

Issue #943, opened 2023-02-09
[SUGGESTION] Implement monetary transactions as a function


LordChunk:
### The problem

It is currently impossible to safely conduct transactions between players. The most common way is to use the ``AddMoney`` and ``RemoveMoney`` functions, but this causes several problems. 

Firstly, it is not possible to see where money comes from. Yes, you do have a ``reason`` parameter, but this is optional.

You can't use SQL transactions to safely update players' balance, as the mutations are running separately.

### Ideal solution

Store player balances separately from the player table in a dedicated table with support for different bank account types. Also have a table for tracking transactions between accounts and other bank account mutations. Also, have a dedicated function for transferring money between different accounts.

### Alternative solutions

Implement a money transfer method. This would not require a dedicated table, which would lead to big migration overhead for existing servers. This transfer method would be able to transfer money of both cash and bank balances to different users, or be able to convert cash to bank balances and vice versa.

PatalJunior:
> You can't use SQL transactions to safely update players' balance, as the mutations are running separately.

I'm confused by your wording here, since I'm not sure if you are speaking about [SQL Transactions](https://www.linode.com/docs/guides/a-primer-on-sql-transactions/) or just a transaction, the act of exchanging something, so I'll propose a implementation assuming you are speaking about a normal transaction.

Upon reading thru the [oxmysql documentation](https://overextended.github.io/docs), there is a method, called a [transaction ](https://overextended.github.io/docs/oxmysql/Usage/transaction/) that only executes at the same time, and if a problem occurs a rollback occurs, so this would be the indicated method to use on said implementation of your idea.

LordChunk:
In your quoted line I'm talking about SQL transactions (not monetary transactions).

I am ware of oxmysql's support for SQL transactions, but you cannot use those if you wish to use the native QB-Core events as these do not allow you to pass a SQL transaction. Fruthermore, events to update a player's balance may also be triggered from the client in which case you can't create an SQL transaction outright.

PR #1099, opened 2024-04-18
Feat: Added money transfer functions

Cocodrulo:
## Description

Implementation of the suggestion [[SUGGESTION] Implement monetary transactions as a function](https://github.com/qbcore-framework/qb-core/issues/943). There have been added to two new functions. One in functions (`QBCore.Functions.TransferMoney`) and other one on player (`Player.Functions.TansferTo`), both aproach the same objective, trasnfering money betwwen two players if both are online, online one is online or both are offline.

## Checklist


MonsieurBibo:
Black money is not in the core, but crypto is, and you didn't put it in ?

Cocodrulo:
Good point, but actually that information is only commented information, I mean is not really whitelisting those monetypes it is only info not actual code.

Anyway I'll fixed it tomorrow

Cocodrulo:
Fixed the annotation and also renamed the `Player.Functions.TransferTo` to `Player.Functions.TransferMoneyTo` to make it self explainable.

server/player.lua
```diff
@@ -337,6 +337,29 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
         return true
     end
 
+    function self.Functions.TransferMoneyTo(emitermoneytype, receivercid, receivermoneytype, quant, reason)
```
Scartane, line 340: `quant` should be replaced by `amount` since it's the wording use by default for this kind of property

server/functions.lua
```diff
@@ -609,3 +609,51 @@ function QBCore.Functions.PrepForSQL(source, data, pattern)
...
+---@param receivercid string
+---@param receivermoneytype string -- Only money types in QBConfig.Money.MoneyTypes
+---@param quant number
+---@param reason string
+---@return boolean
+function QBCore.Functions.TransferMoney(emitercid, emitermoneytype, receivercid, receivermoneytype, quant, reason)
+    local EmiterPlayer = QBCore.Functions.GetPlayerByCitizenId(emitercid)
+    local ReceiverPlayer = QBCore.Functions.GetPlayerByCitizenId(receivercid)
+    if not tonumber(quant) then return false end
+    quant = tonumber(quant) or 0
```
Scartane, line 625: operation with 0 should not be done I think. Probably you can do `if not tonumber(quant) or  tonumber(quant) <= 0 then return false end` or just `if tonumber(quant) <= 0 then return true end`

server/player.lua
```diff
@@ -337,6 +337,29 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
...
+        local ReceiverPlayer = QBCore.Functions.GetPlayerByCitizenId(receivercid)
+        if not tonumber(quant) then return false end
+        quant = tonumber(quant) or 0
```
Scartane, line 343: same remark at the first file regarding 0 transaction handling

server/player.lua
```diff
@@ -337,6 +337,29 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
...
+        local errorOnLast = false
+        if not self.Functions.RemoveMoney(quant, emitermoneytype, reason) then return false end
+        if ReceiverPlayer then
+            if not ReceiverPlayer.Functions.AddMoney(quant, receivermoneytype, reason) then errorOnLast = true end
+        else
+            local result = MySQL.single.await('SELECT money FROM players WHERE citizendid = ?', { receivercid })
+            if not result then errorOnLast = true end
+            result = json.decode(result)
+            result[receivermoneytype] += quant
+            if not MySQL.update.await('UPDATE players SET money = ? WHERE citizenid = ?', { json.encode(result), receivercid }) then return false end
```
Scartane, line 353: instead of `return false` it should be ` errorOnLast = true`

Scartane (commented):
Some thing to change to prevent failure and I think code can be optimize a little bit (wording etc included)

server/functions.lua
```diff
@@ -609,3 +609,51 @@ function QBCore.Functions.PrepForSQL(source, data, pattern)
...
+
+---Do a money transactio between players
+---@param emitercid string
+---@param emitermoneytype string -- Only money types in QBConfig.Money.MoneyTypes
+---@param receivercid string
+---@param receivermoneytype string -- Only money types in QBConfig.Money.MoneyTypes
+---@param quant number
+---@param reason string
+---@return boolean
+function QBCore.Functions.TransferMoney(emitercid, emitermoneytype, receivercid, receivermoneytype, quant, reason)
```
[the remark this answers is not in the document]
Scartane: sourcecid and targetcid wil lbe better than emiter / receiver no ? What do you think ?

Cocodrulo:
Changed those things that you commented, you were right in all of them jajaja

Scartane:
> Changed those things that you commented, you were right in all of them jajaja

Will check that soon and let you know if I find something else

Cocodrulo:
Sorry I added changes by mistake, changes reverted

Scartane (commented):
LGTM 👍 
I don't know if we can add some log, but regarding the code itself i don't see any other mistake :)

Cocodrulo:
Jus added logs and Money Events, a bit chunky the blocks of code but I don't know how to make it smaller.

GhzGarage:
Feel like this functionality could just be added onto `AddMoney` and `RemoveMoney`

Cocodrulo:
I don't understand that

PR #1102, opened 2024-05-16
Update events.lua

TehSteel:
## Description

Added a trigger whenever a player leaves the server.

## Checklist


GhzGarage:
What're you using this for

TehSteel:
You can't directly access a dropped player's data without "complicating" things too much. 
If you want to access a dropped player's data, you need to listen to `onPlayerLoaded` event and save the citizenid inside a table and when `playerDropped` event occurs, you need to check if the citizenid exists. 
Only then can you use `QBCore.Functions.GetOfflinePlayerByCitizenId`.
Even if you use this method, it makes unnecessary SQL queries instead of making a trigger and passing the Player object.

PR #1109, opened 2024-05-25
feat: remove usable item fn after stopped resource

Z3rio:
## Description
This PR aims to fix an error that can be recreated by trying to use an item after the resource which said item callback from is stopped.

Recreation:
- Give yourself a radio
- Stop qb-radio
- Use the radio item

And you'll see the following error:
```
[      script:qb-core] SCRIPT ERROR: citizen:/scripting/lua/scheduler.lua:479: attempt to concatenate a nil value (local 'err')
[      script:qb-core] > UseItem (@qb-inventory/server/functions.lua:167)
[      script:qb-core] > handler (@qb-inventory/server/main.lua:177)
[ script:qb-inventory] SCRIPT ERROR: error object is not a string
```

The way it fixes the error is by saving which resource the "useable item callback" is created by.
And then listening for stopped resources and removing said "useable item callback" upon said resource being stopped.

## Checklist

This needs to be merged alongside:
https://github.com/qbcore-framework/qb-inventory/pull/519


client/functions.lua
```diff
@@ -43,16 +43,13 @@ end
 function QBCore.Functions.LookAtEntity(entity, timeout, speed)
     local involved = GetInvokingResource()
     if not DoesEntityExist(entity) then
-        turnPromise:reject(involved .. ' :^1  Entity does not exist')
```
Mycroft-Studios, line 46: why removing the reject?

Z3rio: `turnPromise` was removed due to the fact that its both undefined and unneeded in this situation. So rejecting here would not work, and just cause an error from trying to index something `nil`.

I guess we should actually `error` or `warn` and return nil instead of returning the error msg

Z3rio: qb-core in its current state:
image
No definition of `turnPromise`.

Z3rio:
> This PR has had 60 days of inactivity & will close within 7 days

no thank you

Qwerty1Verified (approved):

PR #1111, opened 2024-05-26
fix: gang/job validation

Z3rio:
## Description

It has come to my attention that there are ways for the job data to become unsynced.
Such as the `isboss` value being saved as false, even though it's true in the shared files.
(Via no kidding at least 20 tickets of said issue concerning my boss menu and qbcore's `isboss` state)

I'd imagine this can more specifically be recreated via already having a job, and then said jobs shared data being changed, whilst your "locally" saved player job data wouldn't update.

Therefore this PR aims to introduce validation of said data, to make sure it never ever gets out of sync.

## Checklist


ChristianBDev (approved):
Tested LGTM

PR #1119, opened 2024-06-16
Deprecate client-side vehicle spawning function

EinS4ckZwiebeln:
## Description

Deprecate the client-side `SpawnVehicle` function for creating networked vehicles. This function is fine for spawning non-networked vehicles but shouldn't be used for anything else. We should encourage server-sided entity creation through given callbacks.

## Checklist

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -348,7 +348,7 @@ function QBCore.Functions.LoadModel(model)
     end
 end
 
-function QBCore.Functions.SpawnVehicle(model, cb, coords, isnetworked, teleportInto)
+function QBCore.Functions.SpawnVehicle(model, cb, coords, isNetworked, warp)
     local ped = PlayerPedId()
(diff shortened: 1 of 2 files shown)


client/functions.lua
```diff
@@ -357,17 +357,21 @@ function QBCore.Functions.SpawnVehicle(model, cb, coords, isnetworked, teleportI
     else
         coords = GetEntityCoords(ped)
     end
-    isnetworked = isnetworked == nil or isnetworked
+    isNetworked = isNetworked == nil or isNetworked
+    if isNetworked then 
+        print("^3[WARNING] Spawning networked vehicles on the client is deprecated. Please use the server-side callback function instead.^0") 
+    end
```
ChristianBDev, line 363: Could be deleted, just a print check

ChristianBDev (changes requested):
Change the print or delete it and send it back and ill approve

PR #1124, opened 2024-07-09
feat: Add Filter to QBCore export

Jamie9192:
Pass the structure of QBCore, making the value you want equal either "all" or true. Examples in main.lua on client and server side

## Description

Adds ability to filter the core object when importing. Thus reducing memory usage. 
Fixes: importing large object of unneccessary data when not required

## Checklist

- [ X] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [X ] My code fits the style guidelines.
- [X ] My PR fits the contribution guidelines.


ChristianBDev (approved):
Tested LGTM, Good performance, noticeably a difference 
image

Jamie9192:
image_2024-07-11_105110109

Testing shows major improvement only importing Functions. Shared seems to be the main cause of large object. This being client side testing only.

GhzGarage:
@Jamie9192 Could you just add the code to the export with an early return of the full core object if no filters

Jamie9192:
> @Jamie9192 Could you just add the code to the export with an early return of the full core object if no filters

Sure, but the filter function is not _just_ for the core object, and could potentially be used in other resources

Mycroft-Studios (approved):
The only thing i would maybe like to see, is the filter function itself being moved to QBShared, as maybe a:
`QBShared.FilterTable`

Jamie9192:
> The only thing i would maybe like to see, is the filter function itself being moved to QBShared, as maybe a: `QBShared.FilterTable`

This would make sense more so if the core exports were shared, but they are separated, and Shared holds data rather than functions.

As it is a function, I believe its best to be added as such

QBStevie:
.

PR #1133, opened 2024-07-19
Added new functionality: money as an item.

SilantroX:
Added new functionality: money as an item.

## Checklist

- [x] I have personally loaded this code into an updated QBCore project and checked all of its functionality.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -262,12 +262,24 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
     end
 
     function self.Functions.AddMoney(moneytype, amount, reason)
+        source = self.PlayerData.source
         reason = reason or 'unknown'
         moneytype = moneytype:lower()
         amount = tonumber(amount)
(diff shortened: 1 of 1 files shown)


SilantroX (commented):
Modified `player.lua` to use exports for `qb-inventory` and to verify the account to which money is being added. If it's cash, the item named "cash" will be added or removed from the inventory. Additionally, an extra event was added so that when a player joins the server, their money is set to load into their inventory if it wasn't saved previously or if it is the first time using this functionality.

Sanriku-Gaming:
This change is far more likely to be accepted with a Config.MoneyAsItem option, instead of the only option. There are also other events/functions that will need to be changed for this to work.

RojinxChhetri (commented):
I don't think this would be a optimal solution. There must be config option where If developer want cashAsItem = true then use money as Item else current version is better. So I suggest revise it accordingly.

PR #1135, opened 2024-07-20
Feat: Add discord to players DB table

Cocodrulo:
## Description

Nowadays most servers using QBCore uses discord whitelist, so the player will have their discord linked to FiveM. To ease integration with webapges I think discord information in main qbcore database table (players) should be kept to easy locate characters linked to a player.

## Checklist


server/player.lua
```diff
@@ -437,10 +438,11 @@ function QBCore.Player.Save(source)
     local pcoords = GetEntityCoords(ped)
     local PlayerData = QBCore.Players[source].PlayerData
     if PlayerData then
-        MySQL.insert('INSERT INTO players (citizenid, cid, license, name, money, charinfo, job, gang, position, metadata) VALUES (:citizenid, :cid, :license, :name, :money, :charinfo, :job, :gang, :position, :metadata) ON DUPLICATE KEY UPDATE cid = :cid, name = :name, money = :money, charinfo = :charinfo, job = :job, gang = :gang, position = :position, metadata = :metadata', {
+        MySQL.insert('INSERT INTO players (citizenid, cid, license, discord, name, money, charinfo, job, gang, position, metadata) VALUES (:citizenid, :cid, :license, :discord, :name, :money, :charinfo, :job, :gang, :position, :metadata) ON DUPLICATE KEY UPDATE cid = :cid, name = :name, money = :money, charinfo = :charinfo, job = :job, gang = :gang, position = :position, metadata = :metadata', {
```
nick-perry14, line 441: Where is the corresponding Schema update?

nick-perry14 (changes requested):
Not sure how necessary this is, but if you really want it merged you **have** to provide a schema update with the PR.

server/player.lua
```diff
@@ -97,6 +97,7 @@ function QBCore.Player.CheckPlayerData(source, PlayerData)
     if source then
         PlayerData.source = source
         PlayerData.license = PlayerData.license or QBCore.Functions.GetIdentifier(source, 'license')
+        PlayerData.discord = PlayerData.discord or QBCore.Functions.GetIdentifier(source, 'discord'):sub(9)
```
nick-perry14, line 100: Forgot to add.  What is this is null?  calling :sub on a null may cause an issue.  Requires testing.

qbcore.sql
```diff
@@ -3,6 +3,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `citizenid` varchar(50) NOT NULL,
   `cid` int(11) DEFAULT NULL,
   `license` varchar(255) NOT NULL,
+  `discord` varchar(255) NOT NULL,
```
nick-perry14, line 6: This can be null.

nick-perry14 (changes requested):

Cocodrulo:
Just updated, I dont really know if `QBCore.Functions.GetIdentifier` does retrieve null because if we follow the documentation of the function `GetPlayerIdentifierByType` and it is right, it always return a string so, the sub will happen in a string, but change it if the docs are not correct

nick-perry14:
`QBCore.Functions.GetIdentifier` can very much return nil.

image

lioncommunity-es:
You will be correct if you are talking about old qbcore, now it uses the native: 

https://github.com/qbcore-framework/qb-core/blob/e488cbe61e661dc86d16d9e9e7ed4b3e1e341cea/server/functions.lua#L24

Cocodrulo:
You will be correct if you are talking about old qbcore, now it uses the native:

https://github.com/qbcore-framework/qb-core/blob/e488cbe61e661dc86d16d9e9e7ed4b3e1e341cea/server/functions.lua#L24-27

Irishstevie:
Maybe add something to the config.lua to turn this on and off

Cocodrulo:
Well, because the SQL file was change, I feel that it shouldn't be optional

Cocodrulo:
What about this?

Qwerty1Verified:
Could you possibly elaborate on the use case example you mentioned?

Cocodrulo:
> Could you possibly elaborate on the use case example you mentioned?

Yes, for example if u want to get have an admin menu to see all players data (no matters they are online or not) it is always easier, mostly for RP servers, to locate players based on discord Id.

Also if u want to integrate your sv with a discord bot or maybe a Webpage with discord login this will be helpful.

Discord is the main way gaming communities get along right now, and a better integration with this programme providing more critical data storing can become handy.

GhzGarage:
The idea is fine but needs more thought instead of just saving additional information

Cocodrulo:
As I said before, is not only for just saving data, is for future scripts functionality and for easier Web integration

RojinxChhetri:
> I think it really should be added. This will allow new scripts with new innovations

I agree that this addition is essential. It opens the door for creators to develop innovative scripts and introduces opportunities for further enhancements. It would be a valuable contribution to the project.

RojinxChhetri (approved):

PR #1146, opened 2024-10-22
fix for allowing bank accounts to go insanely negative

Mustachedom:
## Description
currently bank is not in the Dont Allow Minus config option and there is no limit to being negative for the RemoveMoney function

this adds a Config option ( -5000 is the preset) and the RemoveMoney function will now check to see if removing money will leave you more than -5000 (or what someone sets it to) and if it will then it returns false, if not and they have enough then it continues with the rest of the function

Yes it fixes numerous exploits 

## Checklist

- [x ] My code fits the style guidelines.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -341,7 +341,7 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
                 end
             end
         end
-        if self.PlayerData.money[moneytype] - amount <= QBConfig.Money.MinusLimit then return false end
+        if self.PlayerData.money[moneytype] - amount < QBConfig.Money.MinusLimit then return false end
         self.PlayerData.money[moneytype] = self.PlayerData.money[moneytype] - amount
 
         if not self.Offline then
--- a/server/player.lua
+++ b/server/player.lua
@@ -341,6 +341,7 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
                 end
             end
         end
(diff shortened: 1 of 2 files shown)


server/player.lua
```diff
@@ -341,6 +341,7 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
                 end
             end
         end
+        if self.PlayerData.money[moneytype] - amount <= QBConfig.Money.MinusLimit then return false end
```
Qwerty1Verified, line 344: Looks good to me, I'd probably just make it `<` rather than `<=` to comply with the wording of 'maximum', as this would only be true for >-4999.

Very minor change though and is just to be consistent with the comment.

Mustachedom: made the change and contributed

Qwerty1Verified: Thank you, I'm sorry to make you add to it again but we tend to use `QBCore.Config` to reference values rather than `QBConfig` just to prevent confusion and keep things consistent with QB resources outside of the core. It's another one of those "Consistency" things, not sure how I missed it the first time. This would just be for `server/player.lua:344` of course.

I do like this feature as it essentially simulates a proper overdraft.

Qwerty1Verified (approved):

server/player.lua
Mustachedom: should be good to go!

PR #1152, opened 2024-12-09
Money as Item

Cocodrulo:
## Description

I don't know why QBCore has never include this feature widely used by RP servers. Some inventories include this feature, but I don't think it is a job for inventory, I believe it is for the core framework of the server. The only thing to take into account is that I'm using a function exported from the qb-inventory, so other inventories should adapt to this change and provide compativility.

Open to suggestions!!

## Checklist

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -7,8 +7,10 @@ QBConfig.StatusInterval = 5000                          -- how often to check hu
 
 QBConfig.Money = {}
 QBConfig.Money.MoneyTypes = { cash = 500, bank = 5000, crypto = 0 } -- type = startamount - Add or remove money types for your server (for ex. blackmoney = 0), remember once added it will not be removed from the database!
+QBConfig.Money.MoneyAsItems = false                                 -- If true the currencies set below will be handled as items, if false money will be handled as normal data in playerdata
+QBConfig.Money.MoneyItems = { cash = 'cash' }                       -- type = item name - Add or remove money items for your server (for ex. blackmoney = 'blackmoney'). Items must be defined in items.lua
 QBConfig.Money.DontAllowMinus = { 'cash', 'crypto' }                -- Money that is not allowed going in minus
-QBConfig.Money.MinusLimit = -5000                                    -- The maximum amount you can be negative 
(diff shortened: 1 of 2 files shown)


Jamie9192 (changes requested):
Like the way you approached having this as a config option, defaulting to false. 

The way the player.lua was implemented, Would be much better broken out to a function to prevent the repeated code.

Cocodrulo:
> Like the way you approached having this as a config option, defaulting to false.
> 
> The way the player.lua was implemented, Would be much better broken out to a function to prevent the repeated code.

Done!! Thanks for the comment

Cocodrulo:
Well I think that if a server wants its money as item, they would not let that account to be in negative terms. But in case it does, it is all handled by the inventory script. It is how this resource handle it what will determine the outcome. I don't know if I have explained myself correctly. But I don't think modern inventories allow negative quantities of items.

Cocodrulo:
I have test this in my server and I've solved some errors, now works fine

robzZ-glitch:
I tested if the infinite error is solved and the error that you couldn't put money in trunks and other inventories and they are solved everything works fine for me too.

ChristianBDev:
Forgot the Item in the qb-core/shared/items.lua as well as the image in the inventory

`cash                         = { name = 'cash', label = 'Money', weight = 0, type = 'item', image = 'cash.png', unique = true, useable = false, shouldClose = false, description = 'A piece of paper that is useful' },`

Cocodrulo:
> Forgot the Item in the qb-core/shared/items.lua as well as the image in the inventory
> 
> `cash = { name = 'cash', label = 'Money', weight = 0, type = 'item', image = 'money.png', unique = true, useable = false, shouldClose = false, description = 'A piece of paper that is useful' },`

Well, since is optional and the default vale is false, I do not think that is necessary

ChristianBDev:
Its optional when they want cash as an item but when they want it they will be coming to the support channels wanting help with their errors since theres no item so add it in and comment out then well make a documentation on it if they want cash as an item, even having it as a item wont do anything if the check for useCashAsAnItem is true.

It's Necessary so add it in so people dont come to support asking us to fix it.

Cocodrulo:
done

PR #1160, opened 2025-01-06
feat: use promises in callbacks + cleanup

Z3rio:
## Description

This PR adds the usage of promises in our callbacks. Meaning we can now trigger callbacks in this manner:

```lua
local response = QBCore.Functions.TriggerCallback("CALLBACK NAME", 123, "test value")
```

Without loosing compatibility with the old format:

```lua
QBCore.Functions.TriggerCallback("CALLBACK NAME", function(response)

end, 123, "test value")
```

This also introduces the `QBCore.Shared.IsFunction` function, to check whether a value is a function **OR** a funcref!

~~Only thing I'm a bit unsure about here, would be the fact that functions become funcrefs, which cant be checked via `type()`, as they return `table` instead of `function`. The solution here is a bit hacky, but the only real "good" way would be the PR [fivem](https://github.com/citizenfx/fivem) and add some kind of function to check whether a value is a funcref or not.~~

## Checklist

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -208,7 +208,9 @@ end)
 
 -- Client Callback
 RegisterNetEvent('QBCore:Client:TriggerClientCallback', function(name, ...)
-    QBCore.Functions.TriggerClientCallback(name, function(...)
+    if not QBCore.ClientCallbacks[name] then return end
+
+    QBCore.ClientCallbacks[name](function(...)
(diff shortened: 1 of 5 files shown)


Z3rio:
This has been **partially** tested by me. Any feedback will be greatly appreciated.
Especially any idea on how to better recognize funcrefs

PR #1161, opened 2025-01-08
Feat: Added Player.Functions.GetName() function and QBCore.Functions.GetName()

Cocodrulo:
## Description

This add two new functions to both client and server side. In the client you can do `QBCore.Functions.GetName()` and this will retrieve a string like this "firstname lastname". Same in the server with `Player.Functions.GetName()`.

## Checklist


Cocodrulo:
Fixed horrific mistake :)

PR #1162, opened 2025-01-09
feat: add GetOnDuty exports

Cocodrulo:
## Description

Added GetOnDuty exports both client and server side. In order to add this `QBCore.Functions.GetOnDuty` was added client-side and a callback was added server-side.

```lua
-- Client side
local cops_func = QBCore.Functions.GetDutyCount('police') -- int
local cops_exp = exports['qb-core']:GetDutyCount('police') -- int

-- Server side
local cops_func = QBCore.Functions.GetDutyCount('police') -- int
local cops_exp = exports['qb-core']:GetDutyCount('police') -- int
```

## Checklist


Mustachedom:
one thing i would add is a check to see if job has a type attached 

so like 

```lua
if QBCore.Shared.Jobs[job].type == nil then 
  check logic for name 
else
   Check logic for type 
 end
 ```
 
 so that way it will pool all people that can do those jobs 
 
 or even a bool as a second param so people can do something like 
 
 ```
 QBCore.Functions.GetOnDuty('police', true)
 and it would check if param2 then check by type assigned to police
 if false or nil then check by name
 ```

Cocodrulo:
Added! (I think)

alfaritsii:
Suggestion: Utilize QBCore:Client:OnPlayerLoaded and QBCore:Client:OnJobUpdate to update the number of duty jobs and store it in the global statebag instead of loop from the table when the function is called, that will make the data can be accessed from both the server side and client side.

Cocodrulo:
Well, idk if thats a better solution, it to messy I think, but i'll look into it?

JericoFX:
the global state bag is a good addition, but is shud make into the core itself so you can return later GlobalState[jobname]

Cocodrulo:
IDK really, I think is messier, but I'll look into it

Cocodrulo:
Added the use of statebags, let me know what you think

PR #1163, opened 2025-01-11
feat: Add function to load animation dictionary

Cocodrulo:
## Description

Added function to load anim dict, there is already one to load anim set, why no this?

Update: Apparently there was QBCore.Functions.RequestAnimDict. The name doesn't fit the other "load" functions that exists in the core and also it does no check if the anim exists or not, so I will stick with the new and and add backward compatibility with the old name.

## Checklist


Cocodrulo:
This should be done in the core

PR #343, opened 2021-11-26
is logged in

ARSSANTO:
**Describe Pull request**
First, make sure you've read and are following the contribution guidelines and style guide and your code reflects that.
Write up a clear and concise description of what your pull request adds or fixes and if it's an added feature explain why you think it should be included in the core.

If your PR is to fix an issue mention that issue here

**Questions (please complete the following information):**
- Have you personally loaded this code into an updated qbcore project and checked all it's functionality?    [yes] 
- Does your code fit the style guidelines?    [yes]
- Does your PR fit the contribution guidelines?     [yes]

PR #1164, opened 2025-01-12
Feature/health armour

Cocodrulo:
## Description

Adds ~~Health~~ and Armour saving and restoring on spawn. Should me merged with [qb-ambulance #343](https://github.com/qbcore-framework/qb-ambulancejob/pull/343)

## Checklist


Cocodrulo:
Set default health to 200

DarkPeluxin:
It worked with armor but did not work with healt.

Cocodrulo:
I've done some investigation and setting health in GTAV is not as easy as setting armour, if you set health to less then a 100, the entity will die, so I think I'll stick with only saving armour

PR #1171, opened 2025-01-29
Add playerID feature to replace source ID

TGIANN:
This feature hides the player's source ID and replaces it with a persistent playerID, ensuring that each player's ID remains the same across sessions. The goal is to use playerID instead of source in QB scripts when performing actions on a player.

For example, replacing `ban source` with `ban playerID` or `giveitem source` with `giveitem playerID`. This ensures that each player has a unique, sequentially increasing ID.

Last Note: Many qb scripts need to be edited to fully utilize this feature

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -28,6 +28,7 @@ function QBCore.Functions.TriggerCallback(name, ...)
     end
 end
 
+---@diagnostic disable-next-line: duplicate-set-field
 function QBCore.Debug(resource, obj, depth)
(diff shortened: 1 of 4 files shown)


TGIANN:
Since I don't use default QB, I have adapted some scripts to be compatible with this new system.

SinisterDevelopment (commented):
the concept works good and the files edited function properly. Just would need to remove the GetPlayer function and only use GetPlayerByPlayerId

Friskky1:
upvote +1 this change is needed for QB tbh

PR #1192, opened 2025-04-10
Improve VehicleInfo Event Handling

DispersiaRoleplay:
Improve VehicleInfo Event Handling with Fallbacks and Optional Key Check

Description:
This PR enhances the QBCore:Client:VehicleInfo event logic for improved compatibility and error safety: Added fallback method to retrieve the vehicle plate using QBCore.Functions.GetPlate() if not directly passed; Added support for checking if the player has vehicle keys only when qb-vehiclekeys is active; Defaulted hasKeys to true to avoid issues when the key system is not used; Cleaned up the logic for packaging and passing vehicle information.

These changes make the system more robust and prevent crashes on servers not using qb-vehiclekeys.

- [yes] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [yes] My code fits the style guidelines.
- [yes] My PR fits the contribution guidelines.

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -160,23 +160,34 @@ RegisterNetEvent('QBCore:Command:DeleteVehicle', function()
 end)
 
 RegisterNetEvent('QBCore:Client:VehicleInfo', function(info)
-    local plate = QBCore.Functions.GetPlate(info.vehicle)
-    local hasKeys = true
+    -- Check if the info object and vehicle data exist
+    if not info or not info.vehicle then return end
 
-    if GetResourceState('qb-vehiclekeys') == 'started' then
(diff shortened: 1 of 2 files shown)


Qwerty1Verified:
Hello, thank you for your PR submission.

Currently in the codebase, plate is never passed through the `info` table, which makes that specific code somewhat redundant. I think I'd also prefer to send the event regardless of whether the plate is present since I'm not sure what the edge cases are there, and where that'd be preventing data from being sent. 

The code does require a plate for the key check, but for that we could avoid the check if a plate isn't present, similar to how the check is avoided if `qb-vehiclekeys` isn't running.

GhzGarage:
This PR has seemed to stray from it's original intended purpose

PR #1200, opened 2025-07-29
Resolved all warning from Lua Diagnostic and added some missing typings

MuDzYNN:
## Description

Basically it won't change any functionality. I just fixed all of warnings that comes from Lua Diagnostic extension. Additionally I added some missing typing for function from `QBCore.Functions` object.

## Checklist

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -37,8 +37,7 @@ RegisterNetEvent('QBCore:Command:GoToMarker', function()
 
     local blipMarker <const> = GetFirstBlipInfoId(8)
     if not DoesBlipExist(blipMarker) then
-        QBCore.Functions.Notify(Lang:t('error.no_waypoint'), 'error', 5000)
-        return 'marker'
+        return QBCore.Functions.Notify(Lang:t('error.no_waypoint'), 'error', 5000)
     end
 
     -- Fade screen to hide how clients get teleported.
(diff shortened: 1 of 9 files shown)


ChristianBDev:
Personally a waste of a PR, most of what you changed is the same output but the first change at the top, why define coords when only calling it 3 times when you can keep it how it is in the first place?

MuDzYNN:
Yeah I know, that's why I'm wrote that this don't change any functionality. It's simply better to work in a project that doesn't have unnecessary warnings. About defining coords - I also do that to resolve warning about missing parameters, because `CreateVehicle` requires 7 arguments and only received 5.

Z3rio:
> Personally a waste of a PR, most of what you changed is the same output but the first change at the top, why define coords when only calling it 3 times when you can keep it how it is in the first place?

I dont really understand this take, especially why its so harsh. Ofcourse this provides value? And its obvious someone put down time and effort to make this....

Opening up a project/resource and seeing a bunch of warnings, most of which arent even "real", isnt ideal. As it defers us from real warnings.
Same with having a (permanently) failing lint test, that just makes this repo look bad (constantly having the red cross), and the code being formatted incorrectly, documented/typed incorrectly and such provides a much worse developer experience.

Furthermore, I do actually believe some of these changes are **required**, as FiveM started forcing lua 5.4 which is much more strict with the return & input values. (Such as using `true` instead of `1`, etc)

Z3rio (approved):
Looks good when reading through it, have not tested this ingame or in my editor, though.

PR #1202, opened 2025-09-07
Added static ID support and option to toggle it in config

itzzzpanda:
Added Static IDs Support . Option to toggle the creation of it.

**Files Changed:**
`config.lua`
`server/commands.lua`
`server/functions.lua`
`server/player.lua`
`server/main.lua`

Much Love

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -20,6 +20,7 @@ QBConfig.Player.Bloodtypes = {
 }
 
 QBConfig.Player.PlayerDefaults = {
+    userId = function() if QBConfig.Server.StaticId.Enabled then return QBCore.Player.CreatePlayerId() end end,
     citizenid = function() return QBCore.Player.CreateCitizenId() end,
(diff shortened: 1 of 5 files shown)


itzzzpanda:
What you are saying is right, I think in this version I did not add a function that I think of after the PR. Basically a function like the one in commands.lua, where based on the choice, it will get you the player based on that value. I am open to pr all qb scripts to use this structure. But wanna know before if it has a chance to be added

Z3rio (changes requested):
Hm, I feel like the whole "CreatePlayerId" functionality and so on is unneeded, aswell as the adding of another "userId" column.
If this is to be implemented, we should definentiely just use the already existing "id" column. As to reduce duplication, improve backwards compatibility, and so on.

This could've been / should be a lot simpler, this feels way too complex for no reason, it just creates unnecessary "breaking points".

PR #1208, opened 2025-11-20
Feature/get shared item

Cocodrulo:
## Description

This pull request introduces a new helper function to the shared codebase to simplify access to shared data. The main change is the addition of a generic `GetShared` function, making it easier and safer to retrieve shared data from various shared namespaces.

Helper function addition:

* Added the `GetShared` function to `shared/main.lua`, allowing retrieval of shared data from specified shared namespaces with safe navigation.

## Checklist

### DIFF
--- a/shared/main.lua
+++ b/shared/main.lua
@@ -181,3 +181,5 @@ QBShared.FemaleNoGloves = {
 function GetShared(namespace, item)
     return QBCore.Shared[namespace]?[item]
 end
+
+exports('GetShared', GetShared)
--- a/shared/main.lua
+++ b/shared/main.lua
@@ -173,3 +173,11 @@ QBShared.FemaleNoGloves = {
     [161] = true,
     [165] = true
 }
+
+--- Get a shared item from a shared field
+--- @param namespace 'Vehicles' | 'VehicleHashes' | 'Items' | 'Gangs' | 'Jobs' | 'Locations' | 'Weapons'
+--- @param item string
+--- @return table
+function GetShared(namespace, item)
+    return QBCore.Shared[namespace][item]
+end
--- a/shared/main.lua
+++ b/shared/main.lua
@@ -179,5 +179,5 @@ QBShared.FemaleNoGloves = {
 --- @param item string
 --- @return table
 function GetShared(namespace, item)
-    return QBCore.Shared[namespace][item]
+    return QBCore.Shared[namespace]?[item]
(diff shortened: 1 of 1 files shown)


Z3rio:
I dont personally see which way this adds any value. As any script using qbcore would anyways have to import the QBCore object, and would therefore already have access to that value - the same way your proposed export does.

I do like the idea though, but for it to actually be effective, we would ideally need to refactor quite a lot of qbcore to work with exports instead, if the endgoal is to not have ot import the core object.

ThatMadCap:
> any script using qbcore would anyways have to import the QBCore object

Changes were made to the core recently that allow framework data to be exposed without needing to call the entire framework object. For instance you can request just shared vehicles, jobs, player functions etc.

As I see it this PR is in line with those recent changes.

Cocodrulo:
> > any script using qbcore would anyways have to import the QBCore object
> 
> Changes were made to the core recently that allow framework data to be exposed without needing to call the entire framework object. For instance you can request just shared vehicles, jobs, player functions etc.
> 
> As I see it this PR is in line with those recent changes.

Yes, I think is a approach that can be use better in dimanic code. The current functions in the core are one for each shared type, here you use a string to select the shared namespace allowing you to choose a shared namespace by code without large if statements (switch-like) to ge the data

Z3rio:
> > any script using qbcore would anyways have to import the QBCore object
> 
> Changes were made to the core recently that allow framework data to be exposed without needing to call the entire framework object. For instance you can request just shared vehicles, jobs, player functions etc.
> 
> As I see it this PR is in line with those recent changes.

Oh, sorry about that then. I will need to look into this. Disgard my statement above entirely.

Thank you for making me aware.

Cocodrulo:
Is this going to be checked at any time?

Z3rio:
@Qwerty1Verified

Qwerty1Verified:
I wish the previous work for adding each `GetSharedX..` export was all condensed into one shared export, and then we could make the `item` parameter optional and allow the shared export to do all the work. Having `GetSharedItems`, `GetSharedVehicles`, etc might be considered more explicit, but `GetShared('Items')` is also effective in my opinion.

Perhaps we can discuss and adjust in the future, since we have it set up that way in other iterations of QB.

Thank you for the PR :)

Qwerty1Verified (approved):

PR #1212, opened 2026-01-10
feat: further exports of player related functionalities

Z3rio:
## Description
The idea of this is mainly from chats in the qbcore discord. I am opening this draft PR so we can all work together on implementing this.

The main goal of the PR is to export main player functions, such as adding items, giving money, and all of those sorts of things.

## Todo/Known issues
- Offline player handling?
> We could perhaps use `GetOfflinePlayer`, but I'm not entirely sure majority of the exports work in that case? This is something we need to test
- Is there a better way than having the in-place `playerFunctions` variable?
- If not (^^^^): Add all of the functions to the `playerFunctions` map
- Test all exports & functionality

## Checklist

### DIFF
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -3,7 +3,7 @@ game 'gta5'
 lua54 'yes'
 author 'Kakarot'
 description 'Core resource for the framework, contains all the core functionality and features'
-version '1.3.0'
+version '1.4.0'
 
 shared_scripts {
     'config.lua',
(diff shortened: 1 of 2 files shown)


Qwerty1Verified:
I think it'd be a good idea for us to standardise a naming scheme for these exported player methods too.

Also, Notify is an example of what I mentioned in the Discord about naming collisions, since we have an exported Notify already, and the player object comes with a Notify. Since the PR contains a table of specific methods, it can just be chosen not to export that one from the player object of course

PR #1214, opened 2026-01-23
Fix health metadata not saving on reconnect

### LINKED ISSUE #1213: [BUG] health metadata not update
### Summary

everytime player relogin it health will become full again, 

### Reproduction

try to make damage and deduct health.
then i try quit or logout and enter server again
my hp is below 50% but when relogin my hp full

### Expected behavior

it should carry its original value of health before they quit.

### Actual behavior

it always full health when player reconnect

### Additional context

_No response_

### Last Updated

just make clean new server

### Custom Resources

ox inventory , ox target, zr scoreboard, 

### Resource Rename

no

AndyBodnar:
## Summary

Fixes player health resetting to full on reconnect by properly saving and restoring health metadata.

## Problem

When players disconnect and reconnect, their health resets to full instead of being preserved. This happens because:
1. No `health` field existed in the default metadata
2. Health was never captured from the client
3. Health was never saved to the database
4. Health was never restored on spawn

## Solution

**config.lua**
- Added `health = 200` to default player metadata

**client/loops.lua**
- Modified `QBCore:UpdatePlayer` to send current health and armor to server

**server/events.lua**
- Modified `QBCore:UpdatePlayer` handler to save received health/armor to metadata
- Modified `playerDropped` handler to capture health/armor before saving on disconnect

**client/events.lua**
- Added health/armor restoration in `OnPlayerLoaded` event

## Testing

Tested using [LuaGround](https://github.com/AndyBodnar/LuaGround) - all metadata persistence tests pass.

## Related Issue

Fixes #1213

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -4,6 +4,21 @@
 RegisterNetEvent('QBCore:Client:OnPlayerLoaded', function()
     ShutdownLoadingScreenNui()
     LocalPlayer.state:set('isLoggedIn', true, false)
+
+    -- Restore saved health and armor after ped is ready
+    CreateThread(function()
+        Wait(1000) -- Wait for ped to be fully spawned
+        local ped = PlayerPedId()
+        local health = QBCore.PlayerData.metadata['health']
+        local armor = QBCore.PlayerData.metadata['armor']
+        if health and health > 0 then
+            SetEntityHealth(ped, health)
+        end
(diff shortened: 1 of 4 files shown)


06875878:
Does this also fix the issue of hunger and thirst metadata not saving on reconnect?

PR #1216, opened 2026-02-19
Improve player system memory usage and performance

CaseIRL:
## Description

Rewrites player.lua to use a class-based prototype pattern with shared methods via __index and weak-keyed private state to reduce memory usage and improve performance at scale.

All existing scripts continue to work without any changes needed, QBCore.Functions.GetPlayer() acts as an API layer. Each GetPlayer call still creates closures for the proxy but these are short-lived rather than permanently stored on the player object still a net improvement over the original.

- player.lua: Virtually a full rewrite to class methods
- functions.lua: New helper added, all player lookups updated
- events.lua: Internal player handling cleaned up

**Mock Object Benchmark**

32 players | Old: 87.99 KB | New: 43.43 KB | Saved: 44.56 KB
64 players | Old: 188.37 KB | New: 86.87 KB | Saved: 101.50 KB
128 players | Old: 376.77 KB | New: 173.77 KB | Saved: 203.00 KB
256 players | Old: 753.64 KB | New: 347.64 KB | Saved: 406.00 KB
512 players | Old: 1507.39 KB | New: 695.39 KB | Saved: 812.00 KB
1024 players | Old: 3014.92 KB | New: 1390.92 KB | Saved: 1624.00 KB
2048 players | Old: 6030.92 KB | New: 2782.92 KB | Saved: 3248.00 KB

## Checklist

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -3,6 +3,53 @@ QBCore.Player_Buckets = {}
 QBCore.Entity_Buckets = {}
 QBCore.UsableItems = {}
 
+--- Builds a proxy object for the player
+--- @param player table The player object to build the proxy for
+--- @return table A proxy object for the player
+function QBCore.Functions.BuildPlayerProxy(player)
+    if not player then return nil end
+
+    local functions = {
+        UpdatePlayerData = function(...) return player:UpdatePlayerData(...) end,
+        SetJob = function(...) return player:SetJob(...) end,
+        SetGang = function(...) return player:SetGang(...) end,
+        SetJobDuty = function(...) return player:SetJobDuty(...) end,
+        SetPlayerData = function(...) return player:SetPlayerData(...) end,
+        SetMetaData = function(...) return player:SetMetaData(...) end,
+        GetMetaData = function(...) return player:GetMetaData(...) end,
+        AddRep = function(...) return player:AddRep(...) end,
+        RemoveRep = function(...) return player:RemoveRep(...) end,
+        GetRep = function(...) return player:GetRep(...) end,
+        AddMoney = function(...) return player:AddMoney(...) end,
+        RemoveMoney = function(...) return player:RemoveMoney(...) end,
+        SetMoney = function(...) return player:SetMoney(...) end,
+        GetMoney = function(...) return player:GetMoney(...) end,
+        Notify = function(...) return player:Notify(...) end,
+        HasItem = function(...) return player:HasItem(...) end,
+        GetName = function(...) return player:GetName(...) end,
+        Save = function(...) return player:Save(...) end,
+        Logout = function(...) return player:Logout(...) end,
+        AddMethod = function(...) return player:AddMethod(...) end,
+        AddField = function(...) return player:AddField(...) end,
+    }
+
+    local result = {
+        PlayerData = player.PlayerData,
+        Offline = player.Offline,
+        Functions = functions
+    }
+
+    for name, fn in pairs(player:GetAllExtraMethods()) do
+        functions[name] = function(...) return fn(player, ...) end
+    end
+
+    for name, value in pairs(player:GetAllExtraFields()) do
+        result[name] = value
+    end
+
+    return result
+end
+
 -- Getters
 -- Get your player first and then trigger a function on them
 -- ex: local player = QBCore.Functions.GetPlayer(source)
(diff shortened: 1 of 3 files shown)


CaseIRL:
I forgot to do the PaycheckInterval in this 

```lua
function PaycheckInterval()
    if not next(QBCore.Players) then
        SetTimeout(QBCore.Config.Money.PayCheckTimeOut * (60 * 1000), PaycheckInterval) -- Prevent paychecks from stopping forever once 0 players
        return
    end
    for _, Player in pairs(QBCore.Players) do
        if not Player then return end
        local payment = QBShared.Jobs[Player.PlayerData.job.name]['grades'][tostring(Player.PlayerData.job.grade.level)].payment
        if not payment then payment = Player.PlayerData.job.payment end
        if Player.PlayerData.job and payment > 0 and (QBShared.Jobs[Player.PlayerData.job.name].offDutyPay or Player.PlayerData.job.onduty) then
            if QBCore.Config.Money.PayCheckSociety then
                local account = exports['qb-banking']:GetAccountBalance(Player.PlayerData.job.name)
                if account ~= 0 then
                    if account < payment then
                        TriggerClientEvent('QBCore:Notify', Player.PlayerData.source, Lang:t('error.company_too_poor'), 'error')
                    else
                        Player:AddMoney('bank', payment, 'paycheck')
                        exports['qb-banking']:RemoveMoney(Player.PlayerData.job.name, payment, 'Employee Paycheck')
                        TriggerClientEvent('QBCore:Notify', Player.PlayerData.source, Lang:t('info.received_paycheck', { value = payment }))
                    end
                else
                    Player:AddMoney('bank', payment, 'paycheck')
                    TriggerClientEvent('QBCore:Notify', Player.PlayerData.source, Lang:t('info.received_paycheck', { value = payment }))
                end
            else
                Player:AddMoney('bank', payment, 'paycheck')
                TriggerClientEvent('QBCore:Notify', Player.PlayerData.source, Lang:t('info.received_paycheck', { value = payment }))
            end
        end
    end
    SetTimeout(QBCore.Config.Money.PayCheckTimeOut * (60 * 1000), PaycheckInterval)
end
```

server/functions.lua
```diff
@@ -3,6 +3,53 @@ QBCore.Player_Buckets = {}
...
+    }
+
+    local result = {
+        PlayerData = player.PlayerData,
+        Offline = player.Offline,
+        Functions = functions
+    }
+
+    for name, fn in pairs(player:GetAllExtraMethods()) do
+        functions[name] = function(...) return fn(player, ...) end
```
Qwerty1Verified: I believe I noticed an issue with this whilst testing. Extra methods are defined by resources that call `AddMethod`, so they typically won't need the player object passed as the first parameter. An example you can test with is `GetItemByName` from `qb-inventory`.

Could you test this out on your end and confirm? Thank you

Qwerty1Verified (changes requested):

server/functions.lua
CaseIRL: For sure, was mostly done pretty quick to be honest as a test if it was even do-able, I am still running the modified version on my dev server atm and i did have to make one or 2 other changes I never pushed back here. Only just noticing this though at 4:50am for me so bare with, i'll take a note of it now, and have a test after sleep ❤️

Qwerty1Verified: No worries. Thank you for working on these changes 😃

CaseIRL: image

Your not wrong, thats something i have updated in my version yes, bare with me I have a few things to do today but ill give it all a proper once over again make sure theres nothing else i forgot about :)

server/player.lua
```diff
@@ -1,11 +1,304 @@
...
+function QBPlayer.new(PlayerData, Offline)
+    local self = setmetatable({}, {
+        __index = function(t, key)
+            local priv = Private[t]
+            if priv and priv.extra_methods[key] then
+                return function(...) return priv.extra_methods[key](t, ...) end
+            end
+            return QBPlayer[key]
+        end
+    })
```
Qwerty1Verified, line 27: I think we could even move the metatable and closure creation out of here, to re-use the same metatable and closures per player instance. It saves extra allocations per player, and saves on garbage collection.

Also, the closure on line 23 may possibly give us the same issue we had previously where it's passing itself into external "extra_methods". From what I can tell, the only path that this would at any time occur on is internal player object calls, which isn't entirely the intended path anyways, and likely wouldn't work with this call flow unless `.` syntax is used.

It may be best to eliminate the closure if there's no other need for it, and just return the handler.

server/player.lua
```diff
@@ -1,11 +1,304 @@
...
+local QBPlayer = {}
+QBPlayer.__index = QBPlayer
+QBPlayer.__metatable = false
+
+function QBPlayer.new(PlayerData, Offline)
+    local self = setmetatable({}, {
+        __index = function(t, key)
+            local priv = Private[t]
+            if priv and priv.extra_methods[key] then
+                return function(...) return priv.extra_methods[key](t, ...) end
```
Qwerty1Verified, line 23:
```suggestion
               return priv.extra_methods[key]
```

Qwerty1Verified: This would allow for the same style of call external to occur internally, rather than having either:
`Player:Test(arg1) - function(player, player, arg1)`
or
`Player.Test(arg1) - function(player, arg1)`

It could be used as:
`Player.Test(arg1) - function(arg1)`

Qwerty1Verified (changes requested):
Apologies for the review gap.

I think with these we could save on even more memory, as well as fix a potential future issue on internal calls via the metamethod.

Feel free to let me know what you find 😄

PR #1218, opened 2026-04-23
feat(player): include previous job and gang data in update events

redutzu:
## Description

Adds previous state tracking to player job and gang updates by passing lastJob and lastGang to server and client events.

This provides downstream resources with access to both the old and new state, improving flexibility for event-driven logic without requiring additional state tracking.

The change is backward compatible, as existing handlers can ignore the extra argument.

Additionally refactors duty handling by moving SetDuty event triggering into SetJobDuty. Previously, these events were only fired when using ToggleDuty; now they are consistently triggered whenever duty is changed via SetJobDuty.

Applies to:
- QBCore:Server:OnJobUpdate
- QBCore:Client:OnJobUpdate
- Gang update equivalents
- QBCore:Server:SetDuty
- QBCore:Client:SetDuty

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -271,10 +271,9 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
     end
 
     function self.Functions.SetJobDuty(onDuty)
-        local lastJob = self.PlayerData.job
         self.PlayerData.job.onduty = not not onDuty
-        TriggerEvent('QBCore:Server:OnJobUpdate', self.PlayerData.source, self.PlayerData.job, lastJob)
-        TriggerClientEvent('QBCore:Client:OnJobUpdate', self.PlayerData.source, self.PlayerData.job, lastJob)
+        TriggerEvent('QBCore:Server:SetDuty', self.PlayerData.source, self.PlayerData.job.onduty)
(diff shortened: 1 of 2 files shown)


server/player.lua
```diff
@@ -270,8 +272,8 @@ function QBCore.Player.CreatePlayer(PlayerData, Offline)
 
     function self.Functions.SetJobDuty(onDuty)
         self.PlayerData.job.onduty = not not onDuty
-        TriggerEvent('QBCore:Server:OnJobUpdate', self.PlayerData.source, self.PlayerData.job)
-        TriggerClientEvent('QBCore:Client:OnJobUpdate', self.PlayerData.source, self.PlayerData.job)
+        TriggerEvent('QBCore:Server:SetDuty', self.PlayerData.source, self.PlayerData.job.onduty)
+        TriggerClientEvent('QBCore:Client:SetDuty', self.PlayerData.source, self.PlayerData.job.onduty)
```
Qwerty1Verified, line 276: This would be slightly breaking for any resources that listened to `OnJobUpdate` for duty updates, although I'm not sure how many would do so.

Listening to `QBCore:ToggleDuty` would've also been fragile.

PR #1219, opened 2026-05-10
Add timeouts and validation to asset loaders

MrNewb:
Enhance several asset-loading helpers in client/functions.lua to avoid indefinite blocking and provide explicit status. LoadModel, RequestAnimDict, LoadAnimSet, and LoadParticleDictionary now accept an optional timeout (default 5000ms), return a boolean load result and the resolved name/hash, and perform existence/validity checks before requesting. LoadModel also accepts string model names (converted via joaat) and logs a warning if the model doesn't exist. These changes make asset loading safer and easier to debug when assets fail to load.

## Description

Adds timeout as optional param.
Adds a Boolean + request return.
Adds failure print where applicable.
Stops infinite loops on model/animset/anim dict/particle requests.
No breaking changes and backwards compatible.

## Checklist

- [X ] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [X ] My code fits the style guidelines.
- [X ] My PR fits the contribution guidelines.

this is kinda fun sometimes

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -356,12 +356,23 @@ end
 
 -- Vehicle
 
-function QBCore.Functions.LoadModel(model)
-    if HasModelLoaded(model) then return end
-    RequestModel(model)
-    while not HasModelLoaded(model) do
+function QBCore.Functions.LoadModel(model, timeout)
+    timeout = timeout or 5000
+    local modelName = model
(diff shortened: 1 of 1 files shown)


MrNewb:
Originally wanted to increment in a count, Stevie suggested game time so went with that.

